PR #6, opened 2021-07-21
Refactor pipelines

jswift-stfc:
The Azure Pipelines have been refactored to conform to modularisation. The large '.azure-pipelines.yml' file has been split into template files and a single 'continuous.yml' file, for CI. After merging, the trigger branches need to be updated to include the master branch.


.azure-pipelines/continuous.yml
```diff
@@ -0,0 +1,144 @@
...
+        - template: templates/test.yml
+          parameters:
+            version: '3.6'
+            arch: 'x64'
+            os: 'ubuntu'
+        - template: templates/test.yml
+          parameters:
+            version: '3.7'
+            arch: 'x64'
+            os: 'ubuntu'
```
trisyoungs, line 48: I feel like we should use a matrix job here (https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?tab=yaml&view=azure-devops&tabs=yaml#multi-job-configuration) to greatly reduce the boilerplate in the yml, but there is every possibility that a lot of time will be lost and head-scratching will be required trying to figure out how it works. The documentation on this kind of stuff isn't exactly overflowing with useful examples...

.azure-pipelines/templates/lint.yml
```diff
@@ -0,0 +1,42 @@
...
+      python -m black --target-version=py38 ./scripts
+      python -m black --target-version=py38 ./gui
+      python -m black --target-version=py38 ./widgets
+    displayName: "Apply black code formatting (PEP8 conformance)"
+
+  - script: |
+      pip install flake8
+      python -m flake8 . --exit-zero
+    condition: succeededOrFailed()
+    displayName: "Static code analysis for PEP8 conformance using flake8"
```
trisyoungs, line 27: Please correct me if I'm not understanding this section - first you format the code according to some standard, then check that the code conforms in a subsequent step?  In Dissolve we are more strict in the sense that we only *check* for consistent formatting within the CI - if your PR contains badly formatted code, the CI abruptly fails.

trisyoungs: Now I've continued reading I see that you generate an artifact containing the re-formatted code. Would it be better to push the responsibility for coding standards on to the developers, as I mentioned above?

gudpy.spec
```diff
@@ -6,8 +6,8 @@ block_cipher = None
 
 a = Analysis(['gui/gudpy.py'],
              pathex=['gudrun_classes/', 'scripts/', 'widgets/', '/home/jared/GudPy/GudPy'],
-             binaries=[],
-             datas=[],
+             binaries=[('Gudrun/bin/gudrun_dcs', '.'), ('Gudrun/bin/libget.so', '.'), ('Gudrun/bin/libGudrunN.so', '.'), ('Gudrun/bin/purge_det', '.')],
```
trisyoungs, line 9: Just a comment on something to bear in mind as we go forward - these binaries only cover the neutron version of the code. There is an X-ray flavour too.

tests/test_gudpy_io.py
```diff
@@ -1,5 +1,6 @@
 import sys, os
 from unittest import TestCase
+import unittest
```
trisyoungs, line 3: Probably my Python ignorance here, but doesn't the `import unittest` line make `from unittest import TestCase` irrelevant, or must the `TestCase` class be explicitly requested?

trisyoungs (approved):
Looks OK.  There are some potential optimisations to be made (see comments) but if you decide to try and implement them you probably want to do this as a separate PR!

.azure-pipelines/continuous.yml
jswift-stfc: I was initially using this method when I was doing everything in one file, I think I know how I can do it here also. I will see if I can implement it.

.azure-pipelines/templates/lint.yml
jswift-stfc: You are probably right, this is not something necessary for the CI - I will remove it. Currently, I allow the CI to continue on the code analysis failing, so I will alter that too, so that it fails if code analysis fails.

gudpy.spec
jswift-stfc: Yes, I do keep forgetting about the X-ray flavour. Since all of the testing I have done so far has been regarding the neutron version. At what point should we begin considering the X-ray flavour, or does it go hand-in-hand with the neutron version?

tests/test_gudpy_io.py
jswift-stfc: The main reason for explicitly importing `TestCase` is for compactness.
For instance, if I just `import unittest`, then when I want to use the `TestCase` class, I will have to refer to it as such: `unittest.TestCase`, whereas with the alternate, I can just refer to the class as `TestCase`. I imported `unittest` for using one of its function decorators for skipping tests `@unittest.skip`. However, better practice would be `from unittest import TestCase, skip`, which is how I will modify it.

gudpy.spec
trisyoungs: Not quite hand-in-hand - more like 2 m social distanced, but walking together! We'll discuss at a later date.

jswift-stfc: Got it. I like the analogy!

PR #11, opened 2021-07-23
CI with Gudrun

jswift-stfc:
- Added a range of tests for attempting to load Gudrun files with missing classes/atrributes.
- Added acquisition of Gudrun binaries to the test and build pipelines.
- Gudrun binaries are packed with executable built by pyinstaller.

### DIFF
--- a/.azure-pipelines/continuous.yml
+++ b/.azure-pipelines/continuous.yml
@@ -6,50 +6,57 @@
 trigger:
   branches:
     include:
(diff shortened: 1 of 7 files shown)


jswift-stfc:
The pull request pipeline failed, as it has not been updated to reflect the updates to the parameters used for building and testing.

trisyoungs:
Need to update the PR pipeline to include the `tag` variable.

PR #13, opened 2021-07-26
Redesign the output structure of gudrun.

jswift-stfc:
This was considerably easier than expected. We simply create (or take a named, existing directory), then change the current working directory to that directory, run gudrun_dcs, and then change back to the initial working directory when gudrun_dcs is finished running. Essentially, we are able to pipe the output files of gudrun_dcs into a named directory, or into a directory named after the first of sample data files.

### DIFF
--- a/gudrun_classes/gudrun_file.py
+++ b/gudrun_classes/gudrun_file.py
@@ -1122,12 +1122,14 @@ class GudrunFile:
         else:
             if not os.path.isfile(outpath):
                 os.mkdir(outpath)    
+        cwd = os.getcwd()
         os.chdir(outpath)
         try:
             result = subprocess.run(['gudrun_dcs', self.path], capture_output=True, text=True)
         except FileNotFoundError:
             gudrun_dcs = sys._MEIPASS + os.sep + 'gudrun_dcs'
             result = subprocess.run([gudrun_dcs, self.path], capture_output=True, text=True)            
+        os.chdir(cwd)
         return result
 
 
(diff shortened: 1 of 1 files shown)


trisyoungs (commented):
This looks like a sensible solution to the issue, but we should ask the question whether continually creating new directories with numerical suffixes in order to not tread on old ones is a behaviour that we want. In particular, this may make it harder to implement the iterative inelasticity subtraction as Gudrun expects to find the msubw01 and msub01 files in the same directory, so these would have to be transferred in from whatever is the "last" directory.

The argument for reusing a single directory named after the first / target data file obviously limits the historic data you can easily access, but it may be better to copy out relevant information (e.g. the mdcs01 and mint01 files) to a GudPy working directory, such that all relevant info can be contained in a single directory and plotted / examined at will.

jswift-stfc:
> This looks like a sensible solution to the issue, but we should ask the question whether continually creating new directories with numerical suffixes in order to not tread on old ones is a behaviour that we want. In particular, this may make it harder to implement the iterative inelasticity subtraction as Gudrun expects to find the msubw01 and msub01 files in the same directory, so these would have to be transferred in from whatever is the "last" directory.
> 
> The argument for reusing a single directory named after the first / target data file obviously limits the historic data you can easily access, but it may be better to copy out relevant information (e.g. the mdcs01 and mint01 files) to a GudPy working directory, such that all relevant info can be contained in a single directory and plotted / examined at will.

I see the problems with this solution. Maybe, we can create a 'live' working directory for GudPy whilst it is running, and as a sort of 'cleanup' task, we can partition everything into relevant directories, which are named after the first/target data file alongside a date/time stamp? I think I need to think about this problem a bit more, maybe we can discuss it at some point?

jswift-stfc:
Closing this for now - however we might come back to this at a later stage.

PR #15, opened 2021-07-27
Format code.

jswift-stfc:
Applied autopep8 to format code to conform to PEP8 standards. Then consulted flake8 outputs to ensure that the entire codebase conforms to PEP8 (see https://www.python.org/dev/peps/pep-0008/). Closes issue #16.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@ bin/
 dist/
 *.xml
 .coverage
+*.gud
+*.txt
+nexus_*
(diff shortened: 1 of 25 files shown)


jswift-stfc:
Pipelines were updated in 02c0a683a18d4995ec989a7bdd98050525d77b7d to stop flake8 (see https://pypi.org/project/flake8/) from forcing a zero-exit value. This means that non-conformance to PEP8 causes CI to fail.

trisyoungs (approved):
Nice and tidy!

PR #18, opened 2021-07-28
Iterate by Tweak factor

### LINKED ISSUE #17: Iteratively tweak the tweak factor
Add functionality to iterate gudrun_dcs for a specified number of iterations, using the suggested tweak factor produced for each sample as the sample tweak factor for that sample in the next iteration. The aim is that the expected DCS level is approximately 100% by the end of the final iteration.

jswift-stfc:
Added functionality for iterating gudrun_dcs by tweaking the tweak factor iteratively. Also added test for regular runs of gudrun_dcs and runs of gudrun_dcs that iterate by tweaking the tweak factor. Closes #17.

### DIFF
--- a/gudrun_classes/gudrun_file.py
+++ b/gudrun_classes/gudrun_file.py
@@ -22,7 +22,6 @@ try:
     from composition import Composition
     from element import Element
     from data_files import DataFiles
-    from gud_file import GudFile
 except ModuleNotFoundError:
     sys.path.insert(1, os.path.join(sys.path[0], "scripts"))
     from scripts.utils import (
@@ -40,7 +39,6 @@ except ModuleNotFoundError:
     from gudrun_classes.normalisation import Normalisation
     from gudrun_classes.sample_background import SampleBackground
     from gudrun_classes.sample import Sample
-    from gudrun_classes.gud_file import GudFile
 
 
 class GudrunFile:
@@ -1520,29 +1518,9 @@ class GudrunFile:
             )
         return result
 
-    def iterateByTweakFactor(self, n):
-        # Iteratively tweak the tweak factor for n iterations.
-
-        for i in range(n):
-
-            # Write out what we currently have,
-            # and run gudrun_dcs on that file.
-            self.write_out()
-            self.dcs(path=self.outpath)
-
-            # Iterate through all samples,
-            # updating their tweak factor from the output of gudrun_dcs.
-            for j, sampleBackground in enumerate(self.sampleBackgrounds):
-                for k, sample in enumerate(sampleBackground.samples):
-                    gud = sample.dataFiles.dataFiles[0].replace(
-                                self.instrument.dataFileType,
-                                "gud"
-                            )
-                    gudFile = GudFile(gud)
-                    tweakFactor = float(gudFile.suggestedTweakFactor.strip())
-                    self.sampleBackgrounds[j].samples[k].sampleTweakFactor = (
-                                                tweakFactor
-                    )
+    def process(self):
+        self.write_out()
+        self.dcs(path=self.outpath)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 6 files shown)


gudrun_classes/gudrun_file.py
```diff
@@ -1501,21 +1504,47 @@ def write_out(self, overwrite=False):
...
             )
         except FileNotFoundError:
             gudrun_dcs = sys._MEIPASS + os.sep + "gudrun_dcs"
             result = subprocess.run(
-                [gudrun_dcs, self.path], capture_output=True, text=True
+                [gudrun_dcs, path], capture_output=True, text=True
             )
         return result
 
+    def iterateByTweakFactor(self, n):
```
trisyoungs, line 1523: I would suggest that its probably better to start adding these "useful" functions into their own separate class(es) (here, `TweakFactorIterator` for instance) which take a `GudrunFile` object as a parameter, and keep `GudrunFile` as pure as possible (just read, write, and run). Otherwise it's going to get very big very quickly!

gudrun_classes/gudrun_file.py
```diff
@@ -1501,21 +1504,47 @@ def write_out(self, overwrite=False):
...
+        # Iteratively tweak the tweak factor for n iterations.
+
+        for i in range(n):
+
+            # Write out what we currently have,
+            # and run gudrun_dcs on that file.
+            self.write_out()
+            self.dcs(path=self.outpath)
```
trisyoungs, line 1531: It's only two lines, I know, but part of me wants this to be encapsulated in a single method (`process()` for instance?) which writes the file and runs the binary.

gudrun_classes/gudrun_file.py
```diff
@@ -1501,21 +1504,47 @@ def write_out(self, overwrite=False):
...
+                for k, sample in enumerate(sampleBackground.samples):
+                    gud = sample.dataFiles.dataFiles[0].replace(
+                                self.instrument.dataFileType,
+                                "gud"
+                            )
+                    gudFile = GudFile(gud)
+                    tweakFactor = float(gudFile.suggestedTweakFactor.strip())
+                    self.sampleBackgrounds[j].samples[k].sampleTweakFactor = (
+                                                tweakFactor
+                    )
```
trisyoungs, line 1545: A likely use-case is that I will only want to adjust tweak factors on a subset of defined samples. In this case, it looks like your loop would adjust the tweak factors of all samples regardless.  Of course, the old GudrunGUI was limited in the sense that it didn't bother objectifying the data it read in, so when it came to things like samples being enabled / disabled it's all contained in the GUI.  You don't have that problem, however, so you can just add a data member to your sample class that encapsulates this kind of information, but doesn't get written out.

trisyoungs (changes requested):

trisyoungs:
Some architectural suggestions to consider, thinking about the bigger picture.

gudrun_classes/gudrun_file.py
jswift-stfc: That sounds like a suitable enhancement. I will add an attribute such as 'adjustTweakFactor', and add it to the Sample class.

trisyoungs: I was thinking more along the lines of a generic "runThisSample" data member.  An "adjustTweakFactor" is too specific - the method handling the running, reading, and setting of the tweak factors is the only place where that knowledge is important.

jswift-stfc:
Added process method to GudrunFile class, and removed iterateByTweakFactor method. (closes #20)
Added runThisSample attribute to sample class. (closes #19)
Created TweakFactorIterator class. (closes #21)

trisyoungs (approved):
:+1:

And I have to say that it is fantastic to finally see the ability to manipulate Gudrun input files in a programmatic way.

jswift-stfc:
> programmatic

I agree! Getting to the exciting stuff! :smile:

jswift-stfc:
Just waiting for the pipeline to run, and I'll merge if everything is okay there.

Issue #22, opened 2021-07-28
Create iterator class for in-elasticity subtractions


jswift-stfc:
A class needs to be created for performing iterations of gudrun_dcs with in-elasticity subtractions.

A suitable name for the class might be `WavelengthSubtractionIterator`.

Some notes on the functionality of the iterator:

- Switch between wavelength and Q scale.
- On wavelength scale, for each sample, always set top hat width to 0, scale to 3 (for wavelength), and the name of file for self scattering to the .mint01 file relevant to the sample.
- On the first iteration on the wavelength scale, turn off subtracting of wavelength-binned data.
- On the Q scale, for each sample, always set the top hat width back to its initial value, scale to 1 (for Q), and the name of file for self scattering to the .msubw01 file relevant to the sample.
- On the first iteration on the Q scale, turn on subtracting of wavelength-binned data - and keep it on for the rest of the iterations.

@trisyoungs could you please confirm if this is OK, and add anything you think I missed?

jswift-stfc:
Also an additional question, for means of clarification - when we run for n iterations, do we perform n iterations on each scale?

trisyoungs:
That looks complete to me.  The only other thing I noticed in my old bash script was that an additional value of `-0.01` was manually-appended to the min, max, stepsize line when performing the wavelength run. I can only assume that this indicates a logarithmic bin width should (must?) be used when binning by wavelength.

trisyoungs:
One iteration comprises a wavelength run followed by a standard Q binning run (`W` | `Q`).  Two iterations would be `W` | `Q` | `W` | `Q`, etc.

jswift-stfc:
> That looks complete to me. The only other thing I noticed in my old bash script was that an additional value of `-0.01` was manually-appended to the min, max, stepsize line when performing the wavelength run. I can only assume that this indicates a logarithmic bin width should (must?) be used when binning by wavelength.

Yes, forgot about that! Thanks for pointing it out.

jswift-stfc:
Should the "runThisSample" attribute that I added be considered when iterating samples?

trisyoungs:
Yes.  I think that we can consider the "runThisSample" to be so generic that it should be respected in all tools / methods / iterations that we add.

PR #24, opened 2021-07-30
Iterate by subtracting wavelength

### LINKED ISSUE #23: Inaccuracy during during testing of third iteration of wavelength subtraction
When using the test data for testing the accuracy of the WavelengthSubtractionIterator, some inaccuracy occurs during only the third and final iteration being tested. The attached image shows the accuracy for all samples across each iteration.
runs

jswift-stfc:
- PurgeFile class added, which we can run purge_det from.
- WavelengthSubtractionIterator class added, for iteratively subtracting wavelengths.
- Tests added and updated.

Closes #23, #22.


gudrun_classes/wavelength_subtraction_iterator.py
```diff
@@ -0,0 +1,107 @@
...
+    def enableLogarithmicBinning(self):
+
+        # -ve to signal logarithmic binning
+        self.gudrunFile.instrument.XScaleRangeStep += (-0.01,)
+
+    def disableLogarithmicBinning(self):
+
+        self.gudrunFile.instrument.XScaleRangeStep = (
+            self.gudrunFile.instrument.XScaleRangeStep[: -1]
+        )
```
trisyoungs, line 19: In the interests of clarity I think that your `XScaleRangeStep` member should be split into the individual components (e.g. `XMin`, `XMax`, `XStep`) with a flag controlling the use of logarithmic binning.

gudrun_classes/wavelength_subtraction_iterator.py
```diff
@@ -0,0 +1,107 @@
...
+        # First iteration
+        if i == 0:
+            self.gudrunFile.instrument.subWavelengthBinnedData = False
+            self.collectTopHatWidths()
+            self.collectQRange()
+        else:
+            self.gudrunFile.instrument.subWavelengthBinnedData = True
+        self.applyWavelengthRanges()
+        self.enableLogarithmicBinning()
+        self.gudrunFile.instrument.scaleSelection = 3
```
trisyoungs, line 88: Similarly, an enumeration (which can then map to the integer values) would make the intent of the scale much clearer.

trisyoungs (changes requested):
Looking good.  Some small suggestions for design changes in the comments.

One general point: I'd like to see more comments in the code - in the present context I'm thinking mainly about your `PurgeFile` `__init__`.  I can follow exactly what you're doing, but a few lines here and there briefly describing blocks of operation would be useful for anyone coming in to look at the code as more of an outsider.

gudrun_classes/wavelength_subtraction_iterator.py
jswift-stfc: Yes, I think this makes sense.

jswift-stfc:
> Looking good. Some small suggestions for design changes in the comments.
> 
> One general point: I'd like to see more comments in the code - in the present context I'm thinking mainly about your `PurgeFile` `__init__`. I can follow exactly what you're doing, but a few lines here and there briefly describing blocks of operation would be useful for anyone coming in to look at the code as more of an outsider.

Yes, you are right. Particularly this file is lacking proper documentation. I think I got a bit carried away!

jswift-stfc:
Not sure what happened exactly with the pipeline. The tests failed claiming that "runThisSample" is not a member of "Sample", yet it is. This was added in this PR. Any idea how to resolve @trisyoungs?

trisyoungs:
Something to do with line 504 in `tests/test_gudpy_io.py` where you pop the `runThisSample` property?

jswift-stfc:
> Something to do with line 504 in `tests/test_gudpy_io.py` where you pop the `runThisSample` property?

Oops.. silly mistake. I assumed it was some GitHub concurrency issue. However, it's because I popped the property from a shallow copy of the dictionary.. Thank you!

tests/test_gudpy_io.py
```diff
@@ -237,9 +242,10 @@ def setUp(self) -> None:
             "outputUnits": 0,
             "powerForBroadening": 0.5,
             "stepSize": 0.03,
-            "analyse": False,
+            "analyse": True,
```
trisyoungs, line 245: Perhaps `include` instead?

gudrun_classes/gudrun_file.py
```diff
@@ -108,9 +110,9 @@ def parseInstrument(self, lines):
             ],
             "channelNosSpikeAnalysis": "Channel numbers",
             "spikeAnalysisAcceptanceFactor": "Spike analysis acceptance",
-            "wavelengthRangeStepSize": ["Wavelength", "range", "step", "size"],
+            "wavelengthMin": ["Wavelength", "range", "step", "size"],
             "NoSmoothsOnMonitor": "smooths on monitor",
-            "XScaleRangeStep": "x-scale",
+            "XMin": "x-scale",
```
trisyoungs, line 115: Not sure what this key is for?

trisyoungs (approved):
Couple of small comments - nothing big!

tests/test_gudpy_io.py
jswift-stfc: That seems reasonable!

gudrun_classes/gudrun_file.py
jswift-stfc: This key is used to "find" the line with the XMin, XMax, XStep values on.  Since they are always on the same line, I found it easier to use XMin as a key, and then extract the XMax and XStep values from the line that XMin was found on.

jswift-stfc:
Will also close #25.

PR #27, opened 2021-08-05
Cleanup codebase / better documentation

### LINKED ISSUE #26: Clean-up codebase
Clean-up the codebase in its current state, refactoring any code deemed to be unreadable/incomprehensible. Modularise code where possible. Ensure integrity of code. Provide descriptive comments throughout the codebase, especially where code is not self explanatory.

jswift-stfc:
Added docstrings for all of the main classes/methods. 
General cleanup of comments etc.
Stopped using .format() for string formatting, refactored to use f-strings instead - they are more accepted, and prettier to look at!

Closes #26.


gudrun_classes/beam.py
```diff
@@ -1,13 +1,51 @@
...
+        list of beam profile values.
+    stepSizeAbsorptionMSNoSlices: tuple(float, float, int)
+        contains step size for absorption,
+        m.s. calculation and number of slices.
+    angularStepForCorrections : int
+        Angular step size to be used in corrections [deg.].
+    incidentBeamEdgesRelCentroid : tuple(float, float, float, float)
+        Positions of edges of incident beam relative to centre of sample.
+    scatteredBeamEdgesRelCentroid : tuple(float, float, float, float)
+        Positions of edges of scattered beam relative to centre of sample.
```
trisyoungs, line 29: This feels like another data member we should represent more explicitly in GudPy - e.g. `leftEdge, `rightEdge`, `topEdge`, `bottomEdge` rather than an array of blind `float`s.

gudrun_classes/composition.py
```diff
@@ -1,21 +1,47 @@
...
+    Class to represent an Atomic Composition.
+
+    ...
+
+    Attributes
+    ----------
+    elements : list
+        List of Element objects present in the composition.
+    type_ : str
+        Type of composition e.g. Sample.
```
trisyoungs, line 13: Something it would be useful to enumerate?

gudrun_classes/container.py
```diff
@@ -11,7 +9,47 @@
...
+    composition : Composition
+        Composition object storing the atomic composition of the container.
+    geometry : str
+        Geometry of the container.
+    thickness : tuple(float, float)
+        Upstream and downstream thickness.
+    angleOfRotationSampleWidth : tuple(float, float)
+        Angle of rotation of the container and its width.
+    densityOfAtoms : str
+        Density of atoms in the container (atoms/Angstrom^3)
```
trisyoungs, line 34: There are two possible units - atoms/Angstrom3 and grams/cm3 - for the density (here and elsewhere).  I assume that Gudrun interprets a negative value as one thing and a positive as the other, but again a separate member in GudPy containing the units would be the way to go.

trisyoungs: This could then just be called `density`.

gudrun_classes/data_files.py
```diff
@@ -1,8 +1,47 @@
...
+    Class to represent a set of data files belonging to an object.
+
+    ...
+
+    Attributes
+    ----------
+    dataFiles : list
+        List of filenames belonging to the object.
+    name : str
+        Name of the parent of the data files, e.g. Sample Background
```
trisyoungs, line 12: Possible enumeration, or are we leaving the door open here for "exotic" groupings of sets of samples together, for instance?

gudrun_classes/gud_file.py
```diff
@@ -3,13 +3,91 @@
...
+        'Name' of the GudFile - this is usually the same as the filename.
+    title : str
+        Title of the run.
+    author : str
+        Author of the run.
+    stamp : str
+        Date and time that the run was completed.
+    densityAcm3 : float
+        Number density of sample (atoms/Angstrom^3).
+    densityAcm3 : float
```
trisyoungs, line 29:
```suggestion
    densityGcm3 : float
```

gudrun_classes/gud_file.py
```diff
@@ -3,13 +3,91 @@
...
+        Path to write to, when not overwriting the initial file.
+    name : str
+        'Name' of the GudFile - this is usually the same as the filename.
+    title : str
+        Title of the run.
+    author : str
+        Author of the run.
+    stamp : str
+        Date and time that the run was completed.
+    densityAcm3 : float
```
trisyoungs, line 27: This data member should be `densityAtomsAngstrom` in your current style, but it might be nicer to rename this and its Gcm3 counterpart to `atomicDensity` and `chemicalDensity` respectively.

gudrun_classes/gud_file.py
```diff
@@ -31,24 +109,41 @@ def __init__(self, path):
...
+
+        # Simple cases, we can just extract the stripped lines.
         self.name = self.contents[0].strip()
         self.title = self.contents[2].strip()
         self.author = self.contents[4].strip()
         self.stamp = self.contents[6].strip()
 
+        # Extract the last item of data from the lines.
         self.densityAcm3 = self.contents[8].split(" ")[-1].strip()
         self.densityGcm3 = self.contents[9].split(" ")[-1].strip()
```
trisyoungs, line 148: See comment above.

gudrun_classes/gudrun_file.py
```diff
@@ -44,40 +44,122 @@
...
+        Parses the attributes of the Instrument from the input lines.
+    parseBeam(lines):
+        Initialises a Beam object and assigns it to the beam attribute.
+        Parses the attributes of the Beam from the input lines.
+    parseNormalisation(lines):
+        Initialises a Normalisation object and assigns it
+        to the normalisation attribute.
+        Parses the attributes of the Normalisation from the input lines.
+    parseSampleBackground(lines):
+        Initialiases a SampleBackground object.
```
trisyoungs, line 82: Typo on `initialiases` here and elsewhere.

gudrun_classes/gudrun_file.py
```diff
@@ -1358,8 +1528,32 @@ def parseContainer(self, lines):
...
+        Only use case is as a helper function during parsing.
+
+        Parameters
+        ----------
+        lines : list
+            List of strings. Each element of the list is a line from the
+            input file.
+        key : str
+            Parsing function to call
+            (INSTRUMENT/BEAM/NORMALISATION/SAMPLE BACKGROUND/SAMPLE/CONTAINER)
```
trisyoungs, line 1544: You're going to think I'm obsessed with enumerations....

gudrun_classes/instrument.py
```diff
@@ -1,13 +1,120 @@
...
+        Acceptance factor for final merge.
+        1.0 indicates all groups are accepted.
+    mergePower : int
+        Power used to set X-weighting for merge.
+    subSingleAtomScattering : bool
+        Should we subtract a background from each group prior to merge?
+    byChannel : int
+        Channel for merge.
+        1 means to use error bars,
+        0 means to use uniform weights.
```
trisyoungs, line 77: Should be three options here (via the GUI): "None?", "By detector", "By Channel". How does this translate to input for `gudrun_dcs`?

gudrun_classes/normalisation.py
```diff
@@ -11,7 +9,55 @@
...
+        DataFiles object storing background data files.
+    composition : Composition
+        Composition object storing the atomic composition of the container.
+    geometry : str
+        Geometry of the container.
+    thickness : tuple(float, float)
+        Upstream and downstream thickness.
+    angleOfRotationSampleWidth : tuple(float, float)
+        Angle of rotation of the container and its width.
+    densityOfAtoms : str
```
trisyoungs, line 35: `density` / `densityUnits`

gudrun_classes/sample.py
```diff
@@ -11,8 +9,83 @@
...
+        TABLES / TRANSMISSION monitor / filename
+    sampleTweakFactor : float
+        Tweak factor for the sample.
+    topHatW : float
+        Width of top hat function for Fourier transform.
+    minRadFT : float
+        Minimum radius for Fourier transform.
+    gor : float
+        Broadening power on g(r).
+        0 = constant, 0.5 = sqrt(r), 1 = r
```
trisyoungs, line 50:
```suggestion
    gor : float
        Broadening of g(r) at r = 1 Angstrom
```
You might consider giving this a better name - `grBroadening` perhaps?

gudrun_classes/tweak_factor_iterator.py
```diff
@@ -5,13 +5,51 @@
 
 
 class TweakFactorIterator():
+    """
+    Class to represent a Tweak Factor Iterator.
+    This class is used for iteratively tweakinng by the tweak factor.
```
trisyoungs, line 10:
```suggestion
    This class is used for iteratively tweaking by the tweak factor.
```

trisyoungs (approved):
Excellent.  A monster PR, but excellent!  All those comments improve readability no end.

I've made some suggestions for typos / renaming vars / using enums (man I loooove enums) but because of the size of the current PR you might want to address those in a separate PR if you choose to act on them.

jswift-stfc:
> Excellent. A monster PR, but excellent! All those comments improve readability no end.
> 
> I've made some suggestions for typos / renaming vars / using enums (man I loooove enums) but because of the size of the current PR you might want to address those in a separate PR if you choose to act on them.

I will resolve the issues which are easily resolved (typos/alterations to comments), and then get working on another PR for the more functional changes!

gudrun_classes/composition.py
jswift-stfc: Yes we can enumerate this, since it is just Sample / Container / Normalisation / Normalisation Background. I'll implement this in my new PR.

gudrun_classes/container.py
jswift-stfc: So you're saying, that we can determine the units based on whether the value is positive or negative?

gudrun_classes/data_files.py
jswift-stfc: This is similar to the Composition, however normally it is the actual name of the parent e.g. CONTAINER N9 etc.
However, the name doesn't seem to be import apart from coming from a readability perspective.
We could use an enumeration here, unless you think we should keep it this way?

gudrun_classes/gudrun_file.py
jswift-stfc: I am beginning to think so..! However, this is a suitable situation for utilising one. I'll make this change in the new PR.

gudrun_classes/instrument.py
jswift-stfc: Aha... I think I have missed something here... Firstly, I think I misinterpreted the Fortran code. Secondly.... you're right, there are three options, and after further consulting the Java code it looks like:
```
0 : "None"
1 : "By detector"
2 : "By channel"
```

This new knowledge throws a bit of a spanner in the works. I had assumed that it was a singular option "By channel?". However, it looks like as you select different options is changes the line in the GudrunFile.
Selecting "None" changes the line to:
`"0          Merge weights: None?"`
Selecting "By detector" changes the line to:
`"1          By detector?"`
Selecting "By channel" changes the line to:
`"2          By channel?"`

Now, I think this is quite the problem, since the actual "key" per se changes depending on the option you have selected. It would make more sense for the "key" to be something like "Merge weights", don't you think? However we will also have to account for people coming in with older files also.

gudrun_classes/sample.py
jswift-stfc: Yes makes sense. I didn't realise this attributes functionality until my trawl through the source code..

gudrun_classes/instrument.py
trisyoungs: True.  However as long as we can write out in the correct current format that `gudrun_dcs` expects, and handle older files on read, then we can just use another, erm, enum to capture the options.

gudrun_classes/data_files.py
trisyoungs: Well.... it makes me think of the bigger picture a bit, now you've said that. Referencing a sample / container / general object by name is a useful thing for us to be able to do.  In principle this should force us to disallow naming one or more objects the same, but for sure this is not the case in GudrunGUI.  I would keep it as is for now, and we'll cross that bridge when (if) we come to it.

gudrun_classes/container.py
trisyoungs: No - I'm saying that the `density` keyword would be a positive, real value that reflected the numerical density, and a separate member `densityUnits` (an enum, or even just a bool) would flag what the units of the density actually are.  Then, when writing (or reading) you know whether to negate the number for the benefit of `gudrun_dcs`.

jswift-stfc: Okay - I think I understand.
`density in gm/cm**3. If -ve, then it is assumed to be in atoms/A**3`

So, if we read the density and it is negative, we save its absolute value, and we make densityUnits 0, for instance 0 = gm/cm^3. Then when we write it back out, we negate it?

gudrun_classes/data_files.py
jswift-stfc: Okay, sounds good!

gudrun_classes/container.py
trisyoungs: Yep, exactly that.  For the purposes of interacting with it in the GUI, we make it easy for ourselves, and suffer the penalty of conversion when we read/write.

jswift-stfc: Cool. Will address this in the new PR.

Issue #33, opened 2021-08-05
Rename density attributes appropiately.


jswift-stfc:
This data member should be `densityAtomsAngstrom` in your current style, but it might be nicer to rename this and its Gcm3 counterpart to `atomicDensity` and `chemicalDensity` respectively.

_Originally posted by @trisyoungs in https://github.com/disorderedmaterials/GudPy/pull/27#discussion_r683497716_

Issue #31, opened 2021-08-05
Fix bug with missing mergeWeights attribute.


jswift-stfc:
Should be three options here (via the GUI): "None?", "By detector", "By Channel". How does this translate to input for `gudrun_dcs`?

_Originally posted by @trisyoungs in https://github.com/disorderedmaterials/GudPy/pull/27#discussion_r683504334_

Add new attribute 'mergeWeights' which references an enum for its values, which can be 0/1/2 meaning None/By detector/By channel. Must be careful with writing out, not to cause any compatibility issues.

Issue #30, opened 2021-08-05
Add densityUnits attribute


jswift-stfc:
There are two possible units - atoms/Angstrom3 and grams/cm3 - for the density (here and elsewhere).  I assume that Gudrun interprets a negative value as one thing and a positive as the other, but again a separate member in GudPy containing the units would be the way to go.

_Originally posted by @trisyoungs in https://github.com/disorderedmaterials/GudPy/pull/27#discussion_r683489964_

Add an additional attribute, to the classes where density is used, for the units of the density (gm/cm^3 or atoms/A^3).

<density in gm/cm**3. If -ve, then it is assumed to be in atoms/A**3>

So, if we read the density and it is negative, we save its absolute value, and we make densityUnits 0, for instance 0 = atoms/A^3. Then when we write it back out, we negate it.

jswift-stfc:
Do we know if this also applies to normalisation/samples? Currently I am assuming that their density is atoms/A^3..

trisyoungs:
Yes.  Safe to assume that, anywhere there's a `density`, there's also an option to specify the units.

Issue #29, opened 2021-08-05
Change "gor" attribute to "grBroadening"


jswift-stfc:
```suggestion
    gor : float
        Broadening of g(r) at r = 1 Angstrom
```
You might consider giving this a better name - `grBroadening` perhaps?

_Originally posted by @trisyoungs in https://github.com/disorderedmaterials/GudPy/pull/27#discussion_r683514072_

PR #35, opened 2021-08-06
Further cleanup of the codebase.

### LINKED ISSUE #28: More explicit attributes, rather than using tuples of floats, individually store those float values.
This feels like another data member we should represent more explicitly in GudPy - e.g. `leftEdge, `rightEdge`, `topEdge`, `bottomEdge` rather than an array of blind `float`s.

_Originally posted by @trisyoungs in https://github.com/disorderedmaterials/GudPy/pull/27#discussion_r683485992_

jswift-stfc:
- Fixed a couple of bugs, particularly the ones regarding merge weights and units of density.
- Split some attributes which were stored as tuples of floats into individual attributes.
- Added script specifically for storing enumerators.
- Enumerator for units of density.
- Renamed some attributes to be more meaningful/descriptive.
- Updated testing suites to reflect the above changes.

Closes #28 #29 #30 #31 #33.

N.B. there are further changes that I still want to make, particularly splitting up more tuples of floats into individual attributes and enumerating where possible. However, I think two smaller PR's is better than one large one.


jswift-stfc:
Forgot to rebase from main for our Windows CI..

gudrun_classes/container.py
```diff
@@ -57,7 +61,8 @@ def __init__(self):
         self.geometry = ""
         self.thickness = (0.0, 0.0)
         self.angleOfRotationSampleWidth = (0.0, 0.0)
-        self.densityOfAtoms = 0.0
+        self.density = 0.0
+        self.densityUnits = 0
```
trisyoungs, line 65: Since we're now using an enum here, I would prefer to see this as:
```suggestion
        self.densityUnits = UnitsOfDensity.ATOMIC
```

gudrun_classes/container.py
```diff
@@ -83,8 +89,19 @@ def __str__(self):
             ''
             )
 
-        return (
+        if self.densityUnits == UnitsOfDensity.ATOMIC.value:
+            units = 'atoms/\u212b^3'
+            density = self.density*-1
```
trisyoungs, line 94: Such a small suggestion, but why not:
```suggestion
            density = -self.density
```

gudrun_classes/enums.py
```diff
@@ -0,0 +1,20 @@
+from enum import Enum
+
+
+class Scales(Enum):
+    Q = 1
+    D_SPACE = 2
```
trisyoungs, line 6: We would commonly refer to this as
```suggestion
    D_SPACING = 2
```

gudrun_classes/gud_file.py
```diff
@@ -24,9 +24,9 @@ class GudFile:
         Author of the run.
     stamp : str
         Date and time that the run was completed.
-    densityAcm3 : float
+    atomicDensity : float
         Number density of sample (atoms/Angstrom^3).
-    densityGcm3 : float
+    chemicalDensity : float
         Number density of sample (g/cm^3).
```
trisyoungs, line 30:
```suggestion
        Chemical density of sample (g/cm^3).
```
"Number density" === "atomic density"

gudrun_classes/gudrun_file.py
```diff
@@ -1448,7 +1520,18 @@ def parseContainer(self, lines):
...
             if i != FORMAT_MAP[key]:
                 FORMAT_MAP[key] = i
-            container.__dict__[key] = float(firstword(lines[FORMAT_MAP[key]]))
+            if key == "density":
+                density = float(firstword(lines[FORMAT_MAP[key]]))
+                if density < 0:
+                    density = abs(density)
+                    container.densityUnits = UnitsOfDensity.ATOMIC.value
+                else:
+                    container.densityUnits = UnitsOfDensity.CHEMICAL.value
```
trisyoungs, line 1529: Presumably these can be written without the `.value`?

gudrun_classes/instrument.py
```diff
@@ -202,6 +204,18 @@ def __str__(self):
             f'0  0  0  0{TAB}0 0 0 0 to end input of specified values\n'
         )
 
+        if self.mergeWeights == MergeWeights.NONE.value:
+            mergeBy = "Merge weights: None?"
+        elif self.mergeWeights == MergeWeights.DETECTOR.value:
+            mergeBy = "By detector?"
+        elif self.mergeWeights == MergeWeights.CHANNEL.value:
+            mergeBy = "By channel?"
```
trisyoungs, line 212: Again, the `.value` seems unnecessary here - could just store the enum member to aid readability and then write out with `.value`.

gudrun_classes/normalisation.py
```diff
@@ -67,7 +71,8 @@ def __init__(self):
         self.geometry = ""
         self.thickness = (0.0, 0.0)
         self.angleOfRotationSampleWidth = (0.0, 0.0)
-        self.densityOfAtoms = 0.0
+        self.density = 0.0
+        self.densityUnits = 0
```
trisyoungs, line 75:
```suggestion
        self.densityUnits = UnitsOfDensity.ATOMIC
```

gudrun_classes/normalisation.py
```diff
@@ -104,6 +109,18 @@ def __str__(self):
             ''
         )
 
+        if self.densityUnits == UnitsOfDensity.ATOMIC.value:
+            units = 'atoms/\u212b^3'
+            density = self.density*-1
+        elif self.density == UnitsOfDensity.CHEMICAL.value:
+            units = 'gm/cm^3'
+            density = self.density
```
trisyoungs, line 117: More `.value`.

gudrun_classes/sample.py
```diff
@@ -93,13 +97,14 @@ def __init__(self):
         self.geometry = ""
         self.thickness = (0.0, 0.0)
         self.angleOfRotationSampleWidth = (0.0, 0.0)
-        self.densityOfAtoms = 0.0
+        self.density = 0.0
+        self.densityUnits = 0
```
trisyoungs, line 101: Assign enum member.

gudrun_classes/sample.py
```diff
@@ -136,6 +141,17 @@ def __str__(self):
             else
             ''
         )
+        if self.densityUnits == UnitsOfDensity.ATOMIC.value:
+            units = 'atoms/\u212b^3'
+            density = self.density*-1
+        elif self.density == UnitsOfDensity.CHEMICAL.value:
+            units = 'gm/cm^3'
+            density = self.density
```
trisyoungs, line 149: `.value`

gudrun_classes/wavelength_subtraction_iterator.py
```diff
@@ -224,10 +217,10 @@ def wavelengthIteration(self, i):
         self.applyWavelengthRanges()
         self.enableLogarithmicBinning()
         self.gudrunFile.instrument.scaleSelection = (
-            self.scales.WAVELENGTH.value
+            Scales.WAVELENGTH.value
```
trisyoungs, line 220: Here and below would read a lot more nicely without the `.value`!

trisyoungs (changes requested):
:+1:, and kepping the enums in a separate file is a sensible way to go.

Almost all of my comments relate to using / comparing your new enums by `type.member.value` when, since we only need to convert to integers when writing out / reading in, readability in the code looks better to me when just using `type.member`, and using `.value` when needing to write an integer value for the benefit of `gudrun_dcs`.

trisyoungs:
Does this need a rebase first?

gudrun_classes/gudrun_file.py
jswift-stfc: Yes, you're right. When writing out we just need to append .value.

gudrun_classes/sample.py
jswift-stfc: > Does this need a rebase first?

Yes it does, I'll sort that out alongside the other issues before merging!

jswift-stfc:
Also need to fix bug introduced in Sample, which is causing the tests to fail.

jswift-stfc:
Closes #36.
Also closing #34 and #32, I decided not to implement these

gudrun_classes/instrument.py
```diff
@@ -142,11 +147,11 @@ def __init__(self):
         self.groupsAcceptanceFactor = 0.0
         self.mergePower = 0
         self.subSingleAtomScattering = False
-        self.byChannel = 0
+        self.mergeWeights = MergeWeights.NONE
```
trisyoungs, line 150: Thinking ahead to "What should the defaults in the new GUI be?" this would be better intialised to `CHANNEL`.

tests/test_gudpy_io.py
```diff
@@ -1165,13 +1233,15 @@ def testLoadMissingContainerAttributesRand(self):
...
-            if key in ["name", "dataFiles", "composition"]:
+            if key in ["name", "dataFiles", "composition", "densityUnits"]:
                 continue
 
             if j == 1:
                 j = 0
             if j >= 2:
                 j += 3
+            if j >= 12:
+                j -= 1
```
trisyoungs, line 1244: For this, and other code like it, can we add some comments to indicate what logic is being applied?

trisyoungs (approved):
Just one minor correction to add to this PR, and two things to consider going forward.

gudrun_classes/container.py
```diff
@@ -30,8 +32,10 @@ class Container:
         Upstream and downstream thickness.
     angleOfRotationSampleWidth : tuple(float, float)
         Angle of rotation of the container and its width.
-    densityOfAtoms : str
-        Density of atoms in the container (atoms/Angstrom^3)
+    density : str
```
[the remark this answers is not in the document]
jswift-stfc: Yeah... I noticed I did this somewhere else and fixed it, thanks for spotting this one!

gudrun_classes/instrument.py
jswift-stfc: I agree, this is something that we should start to consider. Is this something we can discuss at some point?

tests/test_gudpy_io.py
jswift-stfc: Most certainly... Probably should have done this already!

PR #44, opened 2021-08-10
Release CI

jswift-stfc:
Added release pipeline.
Added version to gudpy.spec.

### DIFF
--- a/.azure-pipelines/release.yml
+++ b/.azure-pipelines/release.yml
@@ -6,7 +6,7 @@ trigger:
 pr: none
 
 variables:
-  tag: 2021.1
+  gudrunTag: 2021.1
   targets: >
     {
       linuxPy38:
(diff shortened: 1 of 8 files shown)


jswift-stfc:
Expect PR CI to fail, since I deleted ReleaseNotes.md; since it is specific to `release/*` branches, and here we want to merge into the`master/main` branch.

.azure-pipelines/release.yml
```diff
@@ -0,0 +1,106 @@
+trigger:
+  branches:
+    include:
+      - release/*
+
+pr: none
+
+variables:
+  tag: 2021.1
```
trisyoungs, line 9: For clarity, renaming `tag` to `gudrunTag` would make this more transparent (certainly for me!).

trisyoungs (approved):
All good bar one small suggestion.

.azure-pipelines/release.yml
jswift-stfc: Sounds good! I'll update this across all pipelines.

PR #52, opened 2021-08-17
Build dmg for OSX

### LINKED ISSUE #50: Build DMG for OSX.
Currently, an executable is built for all platforms - it would be useful to bundle the executable into a .dmg for OSX.

jswift-stfc:
Updated gudpy.spec to create .app file, updated build pipeline to create dmg from this.
Fixed various pathing problems that occur when running produced executable/dmg by adding a `main.py`.
Closes #50.


jswift-stfc:
This PR also cleans up the pathing used throughout GudPy. Instead of relative imports throughout scripts, all imports are made from the top level of the package. This means we can stop using 'hacky' fixes such as `pathmagic` and appending paths to `PYTHONPATH`. It is important that after the PR, all branches in active, continuous, development are rebased from the main branch.
This also means that instead of including multiple paths in our executable, we only have to include our top level script, `main.py`, which now acts as an interface for `src/gui/gudpy.py`.

trisyoungs (approved):
:+1:

PR #56, opened 2021-08-20
Added nxsDefinition attribute to Instrument, updated parsing and tests.

### LINKED ISSUE #55: Add nxsDefinitionFile attribute
Since NeXus files have not been considered before, the fact that input files will have a "NeXus definition file", when using .nxs data files, this attribute needs to be added to the `Instrument` class, and needs to be parsed from the input file if it is present (or, only if `dataFileType == nxs`.

jswift-stfc:
Simple PR - simply adds features as described in the title. Closes #55.


trisyoungs (approved):
:+1:

Issue #59, opened 2021-08-23
Parallelisation for tests


jswift-stfc:
Currently, all of the tests, across all agent platforms are run **sequentially** - this leads to long test times, ~30 minutes.
It would be a great improvement to run tests in parallel, across multiple cores. The current test-running framework - which runs the tests, outputs results in JUnit XML format, as wall as code coverage results in Cobertura format, is **nose2**. **nose2** may have the functionality for parallelisation - this should be looked into, alongside the number of cores each agent platform (Windows, Linux, OSX) is able to utilise. However, if we have to move away from **nose2**, it would be paramount that the new test-running framework offers test results and code coverage results in a format accepted by Azure - as this is important to our CI pipelines.

jswift-stfc:
`nose2` has a plugin, `nose2.plugins.mp`, which allows tests to be run in parallel.
```
 LOGICAL_CORES=$(python -c 'import multiprocessing as mp; print(mp.cpu_count())')
 nose2 --with-coverage --coverage-report xml --plugin nose2.plugins.junitxml --junit-xml --plugin nose2.plugins.mp -N $LOGICAL_CORES tests
```

The above code gets the number of logical cores, and then runs nose2, using multiprocessing, with that number of processes.
However this causes a lot of issues - the main reason being that a lot of tests rely on the output files from Gudrun, and after each test we `tearDown`, removing all of the produced files. Since we are running tests in parallel, issues occur where one test finishes and then deletes the files being used by another test. Also, a specific test may produce different results from another, but `gudrun_dcs` **always** overwrites - leading to concurrency issues with the outputs.

We could avoid these problems by using `os.chdir` for each individual test - so each test has its own "working directory". However, currently I am refactoring the tests to accommodate #60 - so I will come back to this at a later point.

jswift-stfc:
Decided not to go forward with this, for now, since #69 halved the testing time on all platforms.

PR #63, opened 2021-08-25
Package binaries better and general CI refactoring

### LINKED ISSUE #62: Package all binaries, rather than specifics
Currently `gudpy.spec` packs specific binaries into the `.exe` it produces. This means that for Windows, the binaries must be suffixed with `.exe`. Also, Windows requires additional libraries from MinGW - which are not released with the binaries for Linux/OSX. Hence, it would probably better to pack the entire directory minus StartupFiles into the .exe.

jswift-stfc:
Initially this PR was just intended to package the Windows-specific MinGW libraries with our executable - this was achieved, but I also extended this into a general CI refactor/cleanup.

- Package MinGW libraries with binaries.
- Modularised/simplified .spec file - no need to specify suffixes or path separators if we include bin/* (minus StartupFiles) as binaries.
- Added zipped version of GudPy for Windows.
- Made use of `conditions`, for in-line bash which is entirely conditional (e.g. building DMG for OSX).
- Display names for in-line bash.
- Split up some in-line bash into separate, more meaningful sections.
- Stop installing Brew - since it comes pre-installed on OSX Azure agents.

Closes #63, #64.


jswift-stfc:
I started working on #59 within this PR, but decided to leave it until a later point. See the comment in #59 for details.

trisyoungs (approved):
:+1:

PR #68, opened 2021-09-02
Run files individually

### LINKED ISSUE #48: Add functionality to run data files individually
The current Java GudrunGUI offers functionality for running data files individually whilst running gudrun_dcs. This is likely done by splitting one Sample with N data files into N samples each with a singular data file.

Should be fairly easy to implement - can be implemented as a class similarly to the TweakFactorIterator  #21 and WavelengthSubtractionIterator #22. A possible name could be `RunIndividualFiles`.

Although - not sure how we should name the samples? We surely cannot have them all share the original name, since the output files of gudrun_dcs would keep getting overwritten?

jswift-stfc:
Created `RunIndividualFiles` class for running data files for each sample individually. The class is simple, and "inherits" (in some sense), some of the methods of `GudrunFile`, for modularity.

The class takes as input a `GudrunFile` object, and partitions it, so each data file for each `Sample` object, has it's own `Sample`.

For example:

```
g = GudrunFile("water.txt") # create GudrunFile object.
runIndividualFiles = RunIndividualFiles(g) # Partition it into individual data files.
runIndividualFiles.process() # Write out, purge, and call gudrun_dcs.
```

Also fixed some issues with docstrings in the iterator classes.

Closes #48.

### DIFF
--- a/src/gudrun_classes/run_individual_files.py
+++ b/src/gudrun_classes/run_individual_files.py
@@ -1,0 +1,176 @@
+from copy import deepcopy
+from src.gudrun_classes.gudrun_file import GudrunFile
+from src.gudrun_classes.purge_file import PurgeFile
+from src.gudrun_classes.data_files import DataFiles
+
+
+class RunIndividualFiles():
+    """
+    Class for running data files individually for samples.
+    Writes out a new instance of a GudrunFile, with samples
+    duplicated for each of their data files.
+
+    ...
+
+    Attributes
+    ----------
+    gudrunFile : GudrunFile
+        GudrunFile to create new instance from, for running
+        data files individually.
+    Methods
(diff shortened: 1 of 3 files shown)


trisyoungs (approved):
Looks OK to me.  The only thing I don't see in there is any check for whether an individual sample is currently selected/activated for processing, but I suppose this is handled by the write routines instead.

jswift-stfc:
> Looks OK to me. The only thing I don't see in there is any check for whether an individual sample is currently selected/activated for processing, but I suppose this is handled by the write routines instead.

Actually I think I should add that here - since in the `TweakFactorIterator` and `WavelengthSubtractionIterator`, we only use samples that are currently selected/activated as you said..

PR #69, opened 2021-09-02
Organise tests

jswift-stfc:
Removed a range of tests, which check that files are produced by running `gudrun_dcs`. Refactor `tearDown`methods for consistency, and modularity, whilst cleaning up. Would like to see how the performance is impacted within the CI pipelines, before any merge is performed.


trisyoungs (approved):
:+1:

PR #71, opened 2021-09-02
Stop indexing lines

### LINKED ISSUE #70: Stop indexing lines
Whilst parsing the input file, rather than indexing lines during assignment as such: `self.instrument.names = lines[0]`, a method such as `getNextToken()` could be implemented, which returns the next line in the input stream. This could improve readability and the general coding style of the parser for `GudrunFiles`.

jswift-stfc:
Instead of indexing lines as such: `instrument.name = firstword(lines[0])`, a stream based approach has been implemented.
The `GudrunFile` class now supports the following methods:

- `getNextToken`
- `peekNextToken`
- `consumeTokens`
- `consumeWhitespace`
- `consumeUpToDelim`

Which are used to maintain a stream within the `GudrunFile` (which is now stored in an attribute, `stream`).

So now, getting the instrument name may look as such: `instrument.name = self.getNextToken()`.

Clearly this is better practice, but also makes the code easier to read - and we no longer have to worry about applying offsets to the indexes after parsing attributes which are of dynamic lengths (`DataFiles`, `Composition`, `groupingParameterPanel` etc.).

This PR closes #70.

### DIFF
--- a/src/gudrun_classes/gud_file.py
+++ b/src/gudrun_classes/gud_file.py
@@ -1,5 +1,6 @@
 import os
 from os.path import isfile
+from src.gudrun_classes.exception import ParserException
 
 
 class GudFile:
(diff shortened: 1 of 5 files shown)


trisyoungs (approved):
Excellent. This cleans things up a whole lot!

Issue #67, opened 2021-09-01
Running working directory


jswift-stfc:
Implement a "running working directory" for GudPy - to offer isolation between the input file and GudPy. Changes to the initial input file are saved to the "running working directory", and the original input file will only become overwritten if specified by the user to do so.
Should be fairly simple to implement - just `os.mkdir` on loading, copy the input file into the running working directory, and then initialise the main `GudrunFile` from that.

PR #72, opened 2021-09-08
Prototype gui

jswift-stfc:
I think now would be a good point for a general check of the GUI-based code, hence I am making this PR now, not to be merged into the `main` branch, but for a check of the current progress, in order to identify any issues, before this section of the codebase becomes even bigger.

I am still experiencing problems with the layouts, unfortunately - this is the main thing that is lacking within the current prototype. I also think that, as we make this prototype functional (beyond saving the entered values), we could possibly think about implementing #67 also.

I understand this is a monsterous PR, but it doesn't necessarily need comprehensive code-review at this stage, this is more to identify that we are on the right track, and where we have veered off.


jswift-stfc:
After some rigorous deciphering of the non-existent docs for promoting widgets to custom PyQt widgets using the designer, I have figured it out! I moved everything in the main window into UI files, and the `GudPyTreeView` is a promoted `QTreeView`. The only "bother" this causes, is that previously the `GudPyTreeView` would be initialized as such: `tree = GudPyTreeView(self, self.gudrunFile)` - the GudrunFile object is then used to create the tree. The GudPyTreeView `__init__` method is called under the hood by `PyUic`, when the UI file is loaded - hence another method `buildTree(self.gudrunFile)` was added which solves this.

jswift-stfc:
Closing this PR for now.

PR #83, opened 2021-09-20
Run samples based on input file.

### LINKED ISSUE #82: Maintain including samples for analysis solely from input file
In our current implementation, we have two attributes regarding including samples in analysis - `include` and `runThisSample`. The former being read from the input file, and the latter being managed exclusively inside GudPy. It's probably a better idea to rename the `include` attribute to `runThisSample`, and remove the `runThisSample` attribute.

jswift-stfc:
Very simple PR that closes #82.

### DIFF
--- a/src/gudrun_classes/sample.py
+++ b/src/gudrun_classes/sample.py
@@ -81,7 +81,7 @@ class Sample:
     stepSize : float
         Step size in radius for final g(r).
     runThisSample : bool
-        Should the sample be runThisSampled in analysis?
+        Should the sample be included in analysis?
     scatteringFraction : float
         Sample environment scattering fraction to compensate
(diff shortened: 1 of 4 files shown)


src/gudrun_classes/sample.py
```diff
@@ -80,8 +80,8 @@ class Sample:
         0 = constant, 0.5 = sqrt(r), 1 = r
     stepSize : float
         Step size in radius for final g(r).
-    include : bool
-        Should the sample be included in analysis?
+    runThisSample : bool
+        Should the sample be runThisSampled in analysis?
```
trisyoungs, line 84: A victim of search/replace!

trisyoungs (approved):
Just one small typo to deal with.

PR #85, opened 2021-09-20
Manage GudPyTreeView using MV architecture solely

### LINKED ISSUE #84: Correct use of MV architecture in GudPyTreeView
Unless I'm misunderstanding the purpose of this implementation (perhaps it's handled differently in Python vs C++) but this seems to go against the purpose of Qt's MV system in the sense that it is explicitly creating `QStandardItems` for entries in the tree, whereas one would typically have a `QAbstractItemModel` that operated on the `GudrunFile` object...

_Originally posted by @trisyoungs in https://github.com/disorderedmaterials/GudPy/pull/75#discussion_r710867543_

jswift-stfc:
- The `GudPyTreeView` has been modified to use a `GudPyTreeModel`, which inherits `QAbstractItemModel`.
- The model does not rely on an external tree to maintain itself, rather it utilises the `index` and `parent` methods.
- No longer need `GudPyNode` class to describe nodes in the tree.
- Implementation is in line with the goal of the `ModelView` architecture.

Closes #84.

### DIFF
--- a/src/gui/widgets/gudpy_tree.py
+++ b/src/gui/widgets/gudpy_tree.py
@@ -216,7 +216,7 @@ class GudPyTreeModel(QAbstractItemModel):
                     0: "Instrument", 1: "Beam",
                     2: "Normalisation", 3: "Sample Background"
                 }
-                return QVariant(dic[index.row()])
+                return QVariant(dic[index.row() if index.row() <= 3 else 3])
             elif isinstance(index.internalPointer(), (Sample, Container)):
                 return QVariant(index.internalPointer().name)
         elif role == Qt.CheckStateRole and self.isSample(index):
--- a/src/gui/widgets/gudpy_tree.py
+++ b/src/gui/widgets/gudpy_tree.py
@@ -8,148 +8,6 @@ from PyQt5.QtCore import (
 )
 
 
-class GudPyNode():
-    """
-    Class to represent a GudPyNode. To be used in a QTreeView.
-
-    ...
-
-    Attributes
-    ----------
-    _data : list
-        Data for node to contain,
-    parent : GudPyNode
-        Parent node.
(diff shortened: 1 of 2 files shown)


src/gui/widgets/gudpy_tree.py
```diff
@@ -213,280 +68,306 @@ def __init__(self, parent, gudrunFile):
...
+            return QModelIndex()
+        # Check that we don't already have the index in reference.
+        if obj not in self.persistentIndexes.keys():
+            # Create the index and add a QPersistentModelIndex
+            # constructed from the index, to the dict.
+            index = self.createIndex(row, 0, obj)
+            self.persistentIndexes[obj] = QPersistentModelIndex(index)
+            return index
+        else:
+            return QModelIndex(self.persistentIndexes[obj])
```
trisyoungs, line 133: Yes - this feels a lot more like how the `index()` functions should be implemented (I think!).

src/gui/widgets/gudpy_tree.py
```diff
@@ -213,280 +68,306 @@ def __init__(self, parent, gudrunFile):
...
-        )
-        return self.createIndex(row, column, child) if child else QModelIndex()
+        -------
+        int
+            Row count.
+        """
+        # If the parent is invalid, then it is a top level node.
+        if not parent.isValid():
+            # Instrument + Beam + Normalisation + N SampleBackgrounds
+            return 3 + len(self.gudrunFile.sampleBackgrounds)
```
trisyoungs, line 285: Here, since we have already talked about combining the `Instrument` and `Beam` tabs, I would rather see a defined numeric constant than the magic number of `3`, in order to make future changes easier and less likely to break this part of the code.

trisyoungs (approved):
MV done proper! Good stuff.

PR #75, opened 2021-09-15
Prototype GUI

jswift-stfc:
Okay.. I think it is time for this monster PR to be opened again.

- Basic GUI structure
- GUI is functional.
- An input file can be loaded, altered, purged, viewed and run.
- UI files are being used for the core of the GUI design.
- Some custom widgets have been implemented e.g. `InstrumentWidget`, `BeamWidget`, etc.
- ModelView architecture is being implemented - and custom `QAbstractTableModels`, `QItemDelegates `and `QTableViews` have been implemented. These can be found in `src/gui/widgets/gudpy_tables.py`.
- Some bug fixes throughout the whole project.

Closes #40.

PR #86, opened 2021-09-22
Instrument Widget UI 

### LINKED ISSUE #78: Instrument Widget Part 1
As a first pass I would make the following sweeping changes and avoid using a grid layout for the entire contents of the `Other` group. 

1) Create a "Monitors" group containing
- Spectrum numbers for incident and transmission beam monitors, perhaps "compressed" into a single horizontal layout as QLabel("Spectrum numbers for beam monitors:") QLabel("Incident") QLineEdit() QLabel("Transmission") QLineEdit(), as you've done for the quiet counts.
- Wavelength range for monitor normalisation
- Quiet count constants
- No of smooths on monitor

2) Create a "Binning" group containing
- All of the X-scale options
- The log binning checkbox
- The Groups parameters panel, contained within a checkable group box (so it can be hidden by default)
- Acceptance factor, merge power, and hard group edges

Leave the rest in the "Other" group for now, and we'll see where we are.

jswift-stfc:
Reorganised the general layout of the `InstrumentWidget`, and split up the fields into further categories (`config/monitors/binning/other`).

Took a more hierarchical approach to layouts - used horizontal layouts inside vertical layouts etc.

Aims to close #78 - however to main aim of this PR is to first ensure that we are on the right track following the feedback of #75.

### DIFF
--- a/src/gui/widgets/instrument_widget.py
+++ b/src/gui/widgets/instrument_widget.py
@@ -1007,6 +1007,8 @@ class InstrumentWidget(QWidget):
         )
 
         # Setup the widgets and slots for the grouping parameter panel.
+
+        self.groupingParameterWidget.setVisible(False)
         self.updateGroupingParameterPanel()
         self.addGroupingParameterButton.clicked.connect(
             self.handleAddGroupingParameter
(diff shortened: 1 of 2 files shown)


trisyoungs (approved):
This looks a *lot* better. Just a couple of suggestions:

1. Perhaps use a sub-group for the spectrum numbers? See screenshot below
2. I'd suggest setting the vertical size policy on the QGroupBoxes to Minimum, and using a vertical spacer to mop-up any extra space (see screenshot again).

iw

PR #93, opened 2021-09-24
Tree context menu

### LINKED ISSUE #89: Context Menu
It would be really nice to have a context menu on the main tree, so that you can right-click on (for instance) a Sample and add a container, "deselect all samples" for analysis (very useful feature) etc...

jswift-stfc:
A context menu has been added so that when right clicking in the tree, a menu appears.
This menu offers the following functionality:
- Copy
- Cut
- Paste
- Insert
- Select All
- Deselect All
- Select only this sample
- Duplicate sample
- Duplicate sample without containers.

Each action is added in context, so depending on the context of the current index selected, different actions are offered.

Closes #89.

### DIFF
--- a/src/gui/widgets/gudpy_tree.py
+++ b/src/gui/widgets/gudpy_tree.py
@@ -1,4 +1,5 @@
-from PyQt5.QtWidgets import QTreeView
+from PyQt5.QtGui import QCursor
+from PyQt5.QtWidgets import QAction, QMenu, QTreeView
 from PyQt5.QtCore import (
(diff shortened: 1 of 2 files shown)


trisyoungs (approved):
:+1: Excellent stuff.

I noticed that if I'm in the middle of a list of SAMPLES and I try to insert a sample background it gets added at the end, so that should really be called "Append Sample Background".  It might be a future enhancement to allow a list of samples to be "split" on to different backgrounds, but probably not necessary now.

PR #94, opened 2021-09-24
Container widget

### LINKED ISSUE #79: Container Widget Part 1
Some things to address:

- There is no layout at all in the "Data Files" group!
- Lots of controls have maximum sizes set (many of the QLabels, and even the main `containerWidget` itself) - this is considered bad practice and will almost certainly break things on other systems.
- The "Total Cross Section Source", "tweak factor", "scattering fraction", and "attenuation coefficient" controls would be better laid out in a 2-column form.
- The radii group has no layout

It's worth mentioning that something missing from many of these layouts is the use of spacers to "mop up" additional space where it is not needed.  A good example of this would be to add a vertical spacer  underneath the `geometryInfoStack_` widget, and which will prevent the big gaps between widgets it contains.

jswift-stfc:
Preliminary PR - to gain feedback and insight.

Ultimately will close #79.

### DIFF
--- a/src/gui/widgets/container_widget.py
+++ b/src/gui/widgets/container_widget.py
@@ -137,7 +137,6 @@ class ContainerWidget(QWidget):
         """
         self.container.geometry = self.geometryComboBox.itemData(index)
         self.geometryInfoStack.setCurrentIndex(self.container.geometry.value)
-        self.geometryInfoStack_.setCurrentIndex(self.container.geometry.value)
 
     def handleUpstreamThicknessChanged(self, value):
         """
@@ -446,7 +445,6 @@ class ContainerWidget(QWidget):
         # Ensure the correct attributes are being
         # shown for the correct geometry.
         self.geometryInfoStack.setCurrentIndex(config.geometry.value)
-        self.geometryInfoStack_.setCurrentIndex(config.geometry.value)
 
         # Setup the widgets and slots for geometry specific attributes.
         # Flatplate
--- a/src/gui/widgets/ui_files/containerWidget.ui
+++ b/src/gui/widgets/ui_files/containerWidget.ui
@@ -32,8 +32,8 @@
        </property>
        <property name="minimumSize">
         <size>
-         <width>291</width>
-         <height>270</height>
+         <width>0</width>
+         <height>0</height>
         </size>
        </property>
        <property name="maximumSize">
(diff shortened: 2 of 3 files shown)


trisyoungs:
OK, so I removed the vertical spacer directly underneath the `dataFilesList` widget, and set the vertical policy of that widget to expanding, and that looks a bit better.  In fact you can remove the vertical spacer in your `configurationGroupBox`'s form layout, since the default behaviour of a form layout is to expand vertically anway.

jswift-stfc:
> OK, so I removed the vertical spacer directly underneath the `dataFilesList` widget, and set the vertical policy of that widget to expanding, and that looks a bit better. In fact you can remove the vertical spacer in your `configurationGroupBox`'s form layout, since the default behaviour of a form layout is to expand vertically anway.

You're right that makes it look a bit better. There is still the considerable empty space inside the  `configurationGroupBox`, but I think it is unavoidable..

trisyoungs:
Ah, sorry - the default style on my Linux box doesn't draw borders on the group boxes, so I hadn't realised that was the issue!  You'll have to lay out the `Configuration` group and a vertical spacer in a vertical layout, then lay out that alongside the `Data Files` group in a horizontal layout

jswift-stfc:
> Ah, sorry - the default style on my Linux box doesn't draw borders on the group boxes, so I hadn't realised that was the issue! You'll have to lay out the `Configuration` group and a vertical spacer in a vertical layout, then lay out that alongside the `Data Files` group in a horizontal layout

That leads us to something like this - still not quite what we are after I think..
Screenshot 2021-09-24 at 13 46 18

I tried with something like this in the designer:

Screenshot 2021-09-24 at 13 56 43

Which ends up like this within the application:

Screenshot 2021-09-24 at 13 58 01

The `dataFilesList` grows too much vertically, even though it's vertical policy is set to minimum, whilst the `containerCompositionTable` shrinks, when it it's vertical policy is set to expanding. Am I missing something here?..

trisyoungs:
Nope - can't explain that, unless you've got a stretch factor set on the `dataFilesList`?  We could probably draw a line under it here and get it merged in if you're more-or-less happy with it?

jswift-stfc:
No stretch factor. Let's merge it for now, and improve in the 2nd pass.

PR #96, opened 2021-09-24
Declutter interface / disable actions if no input file is loaded

### LINKED ISSUE #95: Declutter main window if no input file is loaded
On startup the GUI presents an empty `QTreeView` and an empty `QStackedWidget`, which looks kinda like it isn't working properly.  It might be better to initially hide those widgets, and only show them when an input file is loaded or new input file is created.

jswift-stfc:
If no input file is loaded (i.e when the program initially starts), then the QStackedWidget and GudPyTreeView are hidden, and a range of actions (which require an input file) are disabled until one is loaded.

Closes #95.


trisyoungs (approved):
:+1:

Issue #90, opened 2021-09-23
Object Names and Icons


trisyoungs:
Can we do away completely with the `SAMPLE`, `Sample Background`, and `CONTAINER` names / prefixes on the data objects? Attaching icons illustrating the object type would help to identify things, but even the hierarchical nature of the tree makes this much clearer than it was in the old version.

Also, renaming an object (if relevant) by double-clicking on it is essential.

PR #97, opened 2021-09-24
Normalisation widget

### LINKED ISSUE #80: Normalisation Widget Part 1
Some issues to address:

- Don't set maximum sizes on widgets - in particular the data files and normalisation composition group boxes have maximum widths set.
- Fixing button sizes (e.g. for `addBackgroundDataFileButton`) is again risky practice.  If you want a more controllable button like this consider `QToolButton` with an image as the content. This tends to look better than using the system font for `+` and `-` anyway, IMO.
- Both the data files groups use `QGridLayout`s which seem to prevent the list widgets from expanding vertically (at least in my QtDesigner this is the case).  I would recommend "going manual" and using the more simplistic vertical and horizontal layouts to group together a few widgets at a time, and build up a compound layout in order to get the effect you want. In the present case that would mean putting the +/- buttons along with a vertical spacer in a vertical layout, and laying that out along with the list widget in a horizontal layout. Then you can put the period no label and control in another horizontal layout, and then layout both of these compound layout objects by using a simple vertical layout on the parent group box.

jswift-stfc:
Improved the aesthetics of the NormalisationWidget.

- Stopped setting size constraints on widgets.
- More incremental approach to layouts.
- Generally looks better.
- Haven't addressed the issue with using `QPushButtons` over `QToolButtons` for now, since I'd like to address this is it's own PR. (Possibly a PR alongside #90?)

Closes #80.

### DIFF
--- a/src/gui/widgets/normalisation_widget.py
+++ b/src/gui/widgets/normalisation_widget.py
@@ -282,15 +282,29 @@ class NormalisationWidget(QWidget):
         """
         Slot for handling change to the differential cross section file name.
         Called when a textChanged signal is emitted,
-        from the handleDifferentialCrossSectionFileChanged.
+        from the differentialCrossSectionFileLineEdit.
         Alters the normalisation's differential
         cross section file name as such.
         Parameters
         ----------
         value : str
-            The new value of the handleDifferentialCrossSectionFileChanged.
+            The new value of the differentialCrossSectionFileLineEdit.
         """
         self.normalisation.normalisationDifferentialCrossSectionFile = value
+
+    def handleBrowseDifferentialCrossSectionFile(self):
+        """
+        Slot for handling browsing for a differential cross section file.
+        Called when a cliced signal is emitted,
+        from the browseDifferentialCrossSectionButton.
+        Alters the normalisation's differential
+        cross section file name as such.
+        """
+        filename = QFileDialog.getOpenFileName(
+            self, "Normalisation differential cross section file", ""
+        )
+        if filename[0]:
+            self.differentialCrossSectionFileLineEdit.setText(filename[0])
 
     def handleNormalisationDegreeSmoothingChanged(self, value):
         """
@@ -661,6 +675,10 @@ class NormalisationWidget(QWidget):
         self.differentialCrossSectionFileLineEdit.textChanged.connect(
             self.handleDifferentialCrossSectionFileChanged
         )
+        self.browseDifferentialCrossSectionButton.clicked.connect(
+            self.handleBrowseDifferentialCrossSectionFile
+        )
+
         self.smoothingDegreeSpinBox.setValue(
             self.normalisation.normalisationDegreeSmoothing
         )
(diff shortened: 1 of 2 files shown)


trisyoungs (commented):
Initial comments:

1. Looks like there's an isolated icon on the "Remove" button in the Data Files group - moving all buttons to this, and converting to `QToolButton` will fixup the layouts a lot too, but this could be a separate general PR.
2. The geometry-related stuff should go into a group a la the Container page, and look consistent w.r.t that. To that end I would be tempted to add a sub-group `Thickness` to hold the upstream /downstream (horizontally).

trisyoungs:
Also, I think that the "Total Cross Section Source` option can be included at the bottom of the composition group(s).

trisyoungs (changes requested):
Layout looking good.  One small required addition is the ability to actually choose a "Differential cross section file" - currently the associated "Browse" button does nothing!

src/gui/widgets/normalisation_widget.py
```diff
@@ -282,16 +282,30 @@ def handleDifferentialCrossSectionFileChanged(self, value):
...
+        Called when a cliced signal is emitted,
+        from the browseDifferentialCrossSectionButton.
+        Alters the normalisation's differential
+        cross section file name as such.
+        """
+        filename = QFileDialog.getOpenFileName(
+            self, "Normalisation differential cross section file", ""
+        )
+        if filename[0]:
+            self.differentialCrossSectionFileLineEdit.setText(filename[0])
```
trisyoungs, line 307: Is there a nicer check for the existence of a filename? In C++ I would use `if (!filename.isEmpty())` since the `getOpenFileName` method returns an empty `QString` if the dialog is cancelled.

trisyoungs (approved):
One question about checking for the empty string, then good to go.

src/gui/widgets/normalisation_widget.py
jswift-stfc: Another option could be `if not filename[0] == ""` or `if bool(filename[0])`. General accepted practice for presence check on strings is using `if string`/`if not string`, since empty sequences (strings, in this case) evaluate to false.
I should clarify, that PyQt5 returns `(filename, filter)` from `getOpenFileName` - so `filename[0]` refers to the filename, not the first character in the filename.
I could do something like this:
```
filename, _  = QFileDialog.getOpenFileName(self, "Normalisation differential cross section file", "")
if filename:
     ...
```
Here the filter which is returned is discarded, so the `filename` variable holds ONLY the name of the file.

jswift-stfc: If the dialog is cancelled, in PyQt5, a tuple of two empty strings is returned.

trisyoungs: I think I would go for the explicit capture of the filename from the tuple - that makes things absolutely clear.

jswift-stfc: Got it - I created a new branch `fix-browsing-files`, which I will address this in.

PR #109, opened 2021-09-27
Fix browsing files

### LINKED ISSUE #108: Browsing for Files
There appear to be some inconsistencies in the way files are selected using the various "Browse" / "+" buttons in the UI.

- For "Detector calibration file", "Deadtime Constants File", and "Neutron scattering params file" members on the Instrument page, cancelling the file dialog clears the current text in the `QLineEdit` rather than leaving the original value in place.
- For the "Groups File" on the Instrument page, the "Browse" button doesn't do anything
- For "Incident beam spectrum parameters" on the Beam page, the "Browse" button does nothing.
- For the "+" button on the Data Files groups in the Sample and Container widgets, a file dialog is raised twice.
- There is some odd / undefined behaviour when removing selected Data Files - multiples of the originally-selected file seem to get appended to the list after clicking "-".

jswift-stfc:
Fairly simple PR, fixing up some bugs identified in #108.
Main feature is that across `SampleBackgroundWidgets`, `SampleWidgets` and `ContainerWidgets`, slots are connected to the signals from the widgets in their own method, `setupUI()`, instead of being connected inside `initComponents()` - which caused multiple slots to be connected, where only one should have been.
Also fixed bugs with the `__str__` methods of `DataFiles` and `Sample`.

Closes #108, as mentioned above.

### DIFF
--- a/src/gui/widgets/container_widget.py
+++ b/src/gui/widgets/container_widget.py
@@ -510,7 +510,7 @@ class ContainerWidget(QWidget):
         # Populate cross section source combo box.
         for c in CrossSectionSource:
             self.totalCrossSectionComboBox.addItem(c.name, c)
-    
+
         self.totalCrossSectionComboBox.currentIndexChanged.connect(
             self.handleTotalCrossSectionChanged
         )
@@ -555,7 +555,6 @@ class ContainerWidget(QWidget):
 
         self.angleOfRotationSpinBox.setValue(self.container.angleOfRotation)
         self.sampleWidthSpinBox.setValue(self.container.sampleWidth)
-
 
         # Cylindrical
         self.innerRadiiSpinBox.setValue(self.container.innerRadius)
--- a/src/gui/widgets/container_widget.py
+++ b/src/gui/widgets/container_widget.py
@@ -385,11 +385,13 @@ class ContainerWidget(QWidget):
         regex : str
             Regex-like expression to use for specifying file types.
         """
-        paths = QFileDialog.getOpenFileNames(self, title, ".", regex)
-        for path in paths:
-            if path:
-                target.addItem(path)
+        files = QFileDialog.getOpenFileNames(self, title, ".", regex)[0]
+        for file in files:
+            if file:
+                target.addItem(file.split("/")[-1])
                 self.handleDataFileInserted(target.item(target.count() - 1))
+        if not self.widgetsRefreshing:
+            self.parent.setModified()
 
     def removeFile(self, target, dataFiles):
         """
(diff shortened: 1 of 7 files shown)


src/gui/widgets/container_widget.py
```diff
@@ -384,11 +385,13 @@ def addFiles(self, target, title, regex):
         regex : str
             Regex-like expression to use for specifying file types.
         """
-        paths = QFileDialog.getOpenFileNames(self, title, ".", regex)
-        for path in paths:
-            if path:
-                target.addItem(path)
+        files = QFileDialog.getOpenFileNames(self, title, ".", regex)[0]
```
trisyoungs, line 388: Assuming that this function returns an array of filenames and the selected filter as a tuple, I'd rather see it as the "files, _ = ..." form.

src/gui/widgets/instrument_widget.py
```diff
@@ -656,7 +656,7 @@ def browseFile(self, title, dir=False):
         str[]
         """
         filename = (
-            QFileDialog.getOpenFileName(self, title, "")
+            QFileDialog.getOpenFileName(self, title, "")[0]
```
trisyoungs, line 659: Same here.

src/gui/widgets/sample_background_widget.py
```diff
@@ -116,17 +131,18 @@ def updateDataFilesList(self):
         )
 
     def addFiles(self, target, title, regex):
-        paths = QFileDialog.getOpenFileNames(self, title, ".", regex)
-        for path in paths:
-            if path:
-                target.addItem(path)
+        files = QFileDialog.getOpenFileNames(self, title, ".", regex)[0]
```
trisyoungs, line 134: And here.

src/gui/widgets/sample_widget.py
```diff
@@ -599,10 +602,10 @@ def addFiles(self, target, title, regex):
         regex : str
             Regex-like expression to use for specifying file types.
         """
-        paths = QFileDialog.getOpenFileNames(self, title, ".", regex)
-        for path in paths:
-            if path:
-                target.addItem(path)
+        files = QFileDialog.getOpenFileNames(self, title, ".", regex)[0]
```
trisyoungs, line 605: Here.

trisyoungs (changes requested):
Looking good. Would be nice address the capture of the file(s) from the returned tuple in this PR too, unless there's a compelling reason to do it as a separate piece of work?

trisyoungs (approved):

PR #110, opened 2021-09-27
Icons for Samples/Containers

### LINKED ISSUE #90: Object Names and Icons

jswift-stfc:
Another fairly simple PR - adding a single main feature to close #90.

- Remove SAMPLE/CONTAINER prefixes, when they are present.
- Display an icon beside the object in the `GudPyTreeView`.
- The icons being used currently are mostly for "proof-of-architecture" purposes - these can be switched out for any reasonable/designed icon at a future point.

Closes #90.


jswift-stfc:
I'm not sure why the PR pipeline keeps hanging..

src/gudrun_classes/gudrun_file.py
```diff
@@ -734,7 +734,10 @@ def parseSample(self):
             sample = Sample()
 
             # Extract the sample name, and then discard whitespace lines.
-            sample.name = str(self.getNextToken()[:-2]).strip()
+            sample.name = (
+                str(self.getNextToken()[:-2]).strip()
+                .replace("SAMPLE", "").strip()
+            )
```
trisyoungs, line 740: We should re-add the SAMPLE and CONTAINER prefixes when writing out the file for consistency.

trisyoungs (changes requested):
That neatens things up nicely. I would do two things to the placeholder icons before we commit:

1. Remove the white background from the Sample icon and make it transparent.
2. Make the icons the same size so that they align nicely in the tree

We can beautify the resources at a later date.

trisyoungs:
> I'm not sure why the PR pipeline keeps hanging..

Blame the infrastructure!  May as well cancel and re-run it from Azure...

PR #112, opened 2021-09-28
Toggle grouping parameter panel

jswift-stfc:
Implemented toggling the visibility of the grouping parameter panel within the `InstrumentWidget` - also resolved some layout problems with maximum size constraints on widgets.

One issue is present, that making the grouping parameter panel enlarges all of the group boxes, and then hiding the grouping parameter panel again does not revert these changes.


trisyoungs (approved):
I think we can merge this in as-is - I don't get the same issue on collapsing the group as you mention here - but bearing in mind that we reorganise this widget semi-significantly in upcoming PRs.

PR #122, opened 2021-09-29
Tabify SampleWidget

### LINKED ISSUE #114: Tabify SampleWidget
To make space for plots to be displayed underneath the `SampleWidget`, separate the `QGroupBoxes` into tabs.

jswift-stfc:
The `SampleWidget` has been reorganised to utilise a `QTabWidget` to view the Sample's configuration, composition / geometry, Fourier transform parameters and advanced parameters.
Also modified the way that the **ancestor** widget is found (`GudPyMainWindow`).

Closes #114.

### DIFF
--- a/src/gui/widgets/gudpy_tables.py
+++ b/src/gui/widgets/gudpy_tables.py
@@ -11,7 +11,7 @@ from PyQt5.QtWidgets import (
 from PyQt5.QtGui import QCursor
 
 from src.gudrun_classes.element import Element
-
+from src.gui.widgets.main_window import GudPyMainWindow
 
 class GudPyTableModel(QAbstractTableModel):
     """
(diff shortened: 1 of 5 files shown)


trisyoungs (commented):
My immediate reaction to this change is that it's too many tabs.  I think that a better layout would be to have just two tabs - "<Sample Name>" and "Advanced Options" - since the config, composition / geometry, and Fourier transform options are all common usage, and putting them in separate tabs makes for a slightly clunky UX.

This also brings to mind something related to the graphing - we currently have a reserved space on the right in which we were thinking of putting the main graphing tools in, but it may be a better approach to have a "super tab widget" at the very top level which contains two pages - one with your current stack of widgets, and the other with the graphing tools.  This gets around issues related to size, and gives a full-screen graphing experience to the user.

trisyoungs (approved):
Yes, this feels a lot better.  It also immediately screams of having multiple main tabs, so instead of just "Plots" we can have "MDCS / MINT", "G(r)", "D(r)", "Diagnostic" etc.  Anyways, that can be left for a future PR (when we actually have some graphing!)

PR #126, opened 2021-09-30
Minimal Iteration Dialog

jswift-stfc:
A minimal dialog to enable users to perform iterations has been implemented, currently this will act as a stub for implementing iterating in the future.


trisyoungs (approved):
Good, lays the foundation for all the other fabulous features we'll implement!

One thing - the minimum value for the number of iterations spin can be 1.

Issue #132, opened 2021-10-01
Use QProcess for iterating


jswift-stfc:
Similar to the way `gudrun_dcs` and `purge_det` is being run, the `WavelengthSubtractionIterator` and `TweakFactorIterator` should be modified to utilise `QProcess` when they are being run through the GUI.

Issue #131, opened 2021-10-01
Track progress of purge_det


jswift-stfc:
Similar to that which has been implemented for `gudrun_dcs`, the `progressBar` should track the progress of `purge_det`.

Issue #127, opened 2021-09-30
Reduce automation in running


jswift-stfc:
GudPy calls `purge_det` before `gudrun_dcs` always. It would be better to keep these separate, and inform the user if they haven't ran `purge_det`.

PR #133, opened 2021-10-01
Run related fixes / enhancements / various bug fixes

### LINKED ISSUE #115: Run binaries in separate threads?
Running `purge_det` or `gudrun_dcs` from inside the GUI will make the GUI unresponsive, since it will wait for the response of either script. We could possibly run the binaries in a separate thread or process, in order to keep the GUI responsive during processing? Although this may not be necessary, if it is not expected that someone may keep using the GUI whilst they are processing some data.

### LINKED ISSUE #128: Attributes that affect purging
I am adding a flag to the `GudrunFile` class to determine whether the detectors have been purged or not. However, we must consider that the detectors may be purged, then attributes changed, and it will still look as if it has been purged - although I'm assuming that some of the attributes related to the `GudrunFile` affect the purging process, so it would have to be run again.

As far as I am aware, alterations to the following attributes would require purging again:

- Any configuration/data files, and period numbers.
- Instrument name.
- Spectrum numbers for incident beam monitor.
- Channel numbers for spike analysis and acceptance factor.

Can you confirm that this is correct, and that it would be reasonable to set the flag to `False`, if any of these are changed?

jswift-stfc:
PR to fix up some general issues, whilst fixing run relating issues and enhancing running binaries from the GUI.

- `purge_det` and `gudrun_dcs` are ran using `QProcess` (see #132 regarding iterating). Closes #115.
- Warnings generated when `gudrun_dcs' is attempted to be ran without first purging detectors. Closes #128, #127.
- Dialog for purging - to alter parameters.
- Functionality for excluding Samples and Containers data files from `purge_det`.
- Status bar showing current task and it's progress, in a progress bar - works for `gudrun_dcs` (see #131 regarding `purge_det`).
- Disable controls whilst `gudrun_dcs` and `purge_det` are being run - for safety.
- General fixes of bugs etc.


jswift-stfc:
Closing for now, as I want to make more changes.

src/gui/widgets/iteration_dialog.py
```diff
@@ -102,17 +105,35 @@ def iterate(self):
         if self.tweakValues:
             if self.tweak == Tweakables.TWEAK_FACTOR:
```
trisyoungs, line 106: Just a thought, but there is an issue already in to handle renaming of the good old "tweak factor" - in a similar spirit would the `Tweakables` enum not be better named as `Iterables` or similar?

src/gui/widgets/main_window.py
```diff
@@ -322,3 +405,138 @@ def setUnModified(self):
...
+        self.cut.setEnabled(True)
+        self.paste.setEnabled(True)
+        self.runPurge.setEnabled(True)
+        self.runGudrun.setEnabled(True)
+        self.iterateGudrun.setEnabled(True)
+        self.viewLiveInputFile.setEnabled(True)
+        self.save.setEnabled(True)
+        self.saveAs.setEnabled(True)
+        self.loadInputFile.setEnabled(True)
+        self.loadConfiguration.setEnabled(True)
```
trisyoungs, line 462: This could be contracted into a single function `setControlsEnabled` taking a boolean specifying the enabled state. Then you only have to maintain a single list of controls.

src/gui/widgets/main_window.py
```diff
@@ -322,3 +405,138 @@ def setUnModified(self):
...
+        self.progressBar.setValue(progress if progress <= 100 else 100)
+
+    def progressPurge(self):
+        data = self.proc.readAllStandardOutput()
+        stdout = bytes(data).decode("utf8")
+        print(stdout)
+        if "Total run time" in stdout:
+            QMessageBox.warning(
+                self, "GudPy Warning",
+                f"{nthint(stdout, 0)} detectors made it through the purge."
```
trisyoungs, line 524: Best to add an issue to extend this to instrument specific values for the allowable number of detectors making it through the purge.

src/gui/widgets/sample_background_widget.py
```diff
@@ -0,0 +1,156 @@
...
+        Gives the focus of the SampleBackgroundWidget to the sample.
+        Calls the initComponents method, to load the UI file and populate data.
+        Parameters
+        ----------
+        sampleBackground : SampleBackground
+            SampleBackground object belonging to the GudrunFile.
+        """
+        self.sampleBackground = sampleBackground
+        # Acquire the lock
+        self.widgetsRefreshing = True
```
trisyoungs, line 55: Redundant as this is performed within `initComponents`.

src/gui/widgets/sample_slots.py
```diff
@@ -319,6 +319,7 @@ def handlePeriodNoChanged(self, value):
         self.sample.periodNo = value
         if not self.widgetsRefreshing:
             self.parent.setModified()
+            self.parent.gudrunFile.purged = False
```
trisyoungs, line 322: Same here (account for purge targets?)

trisyoungs (approved):
:+1: Few little bits to consider or capture before merging.

src/gui/widgets/container_slots.py
```diff
@@ -223,6 +223,7 @@ def handlePeriodNoChanged(self, value):
         self.container.periodNo = value
         if not self.widgetsRefreshing:
             self.parent.setModified()
+            self.parent.gudrunFile.purged = False
```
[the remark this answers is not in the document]
trisyoungs: Is `PurgeFile` easily accessible globally? A couple of simple methods in that class would handle it neatly -`notifySampleOrCanChanged()` and `notifyNormalisationOrBackgroundChanged()` for example, of which the former can check the state of `excludeSampleCan` and decide how to act.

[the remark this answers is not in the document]
jswift-stfc: Currently PurgeFile only gets instantiated inside the dialog. Maybe it's better to have a PurgeFile as an attribute associated with the GudrunFile objects?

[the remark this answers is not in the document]
trisyoungs: Yes, I think the relationship between `GudrunFile` and `PurgeFile` is tight enough to warrant the latter being an attribute of the former, and it will solve the present issue.

PR #135, opened 2021-10-05
File related fixes

### LINKED ISSUE #124: Check files exist
Along the same lines as #120, it would be good to have a feature, similarly to that of the current Gudrun Java GUI, for checking that all of the files being used within the run actually exist.

jswift-stfc:
Cleaned up some hinderances in the way that we decide on paths - trying to use absolute paths where possible.
Added a class `GudPyFileSystem`, which keeps track of what files are present in the Gudrun file - and to check if they exist. This can be done through the GUI.

Closes #124.


src/gudrun_classes/fs.py
```diff
@@ -0,0 +1,128 @@
+import os
+from src.gudrun_classes.enums import CrossSectionSource
+
+
+class GudPyFileSystem():
```
trisyoungs, line 5: Is this the best name for the class?  I see "file system" and have immediate thoughts of "What? He's implemented a file system!?"  Maybe `GudPyFileLibrary` would be more appropriate?

src/gudrun_classes/instrument.py
```diff
@@ -116,7 +117,7 @@ def __init__(self):
         None
         """
         self.name = Instruments.NIMROD
-        self.GudrunInputFileDir = ""
+        self.GudrunInputFileDir = os.path.sep
```
trisyoungs, line 120: If we're doing this to spoof the ability to use absolute file names then we should write a comment explaining the reasoning.

trisyoungs (approved):
Yup!  :+1:

Small comments for consideration.

Issue #142, opened 2021-10-11
Graphing


jswift-stfc:
Use `QChart` to plot results of `gudrun_dcs`. It would first be good to do some prototyping - possibly beginning with a `Structure factor` graph, which plots the mint and mdcs files as well as the dcs level.

PR #144, opened 2021-10-12
Basic plotting / bug fixes

jswift-stfc:
The main aim of this PR is to set a baseline for plotting in future. This PR implements basic plotting for the structure factor (mint01, mdcs01) within the GUI. Working towards #142. Also fixed a couple of small bugs here, particularly with the Q range.

### DIFF
--- a/tests/test_gudpy_io.py
+++ b/tests/test_gudpy_io.py
@@ -38,7 +38,6 @@ class TestGudPyIO(TestCase):
                 + "/"
                 + path
             )
-
         self.expectedInstrument = {
             "name": Instruments.NIMROD,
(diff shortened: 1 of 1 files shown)


trisyoungs (changes requested):
Plotting at last!  Excellent example of where we're heading.

Some things to fixup in this PR before merging:
1. Don't display the plot of all structure factors on the Sample tab - move that to the main `Plots` tab
2. Once moved, put a mdor/mgor (Radial Distribution Functions) plot underneath the Structure Factors graph.
3. Always display the plots, even if no data is available.
4. On loading an input file in a directory where the were existing mint/mdcs files, these were not loaded automatically.

There are other bits and pieces which I'll capture as separate issues.

PR #149, opened 2021-10-13
Iterate GUI

### LINKED ISSUE #132: Use QProcess for iterating

jswift-stfc:
This PR closes #132. It implements iterating properly through the GUI, and also fixes the problem with the iteration dialog not showing properly. The progress bar is basic for now, and will be addressed later on.


trisyoungs (approved):
Now it works!  One small bug to fix before merging - need to refresh the currently-displayed stack page after the iterations are complete. If a Sample is being displayed in the GUI and you tweak the tweak factors, the old tweak factor is displayed until you change between samples.

jswift-stfc:
> Now it works! One small bug to fix before merging - need to refresh the currently-displayed stack page after the iterations are complete. If a Sample is being displayed in the GUI and you tweak the tweak factors, the old tweak factor is displayed until you change between samples.

Ah I didn't even think about that! I'll sort it out now!

Issue #148, opened 2021-10-13
Show results of gudrun_dcs in GUI


jswift-stfc:
Based on the survey, it would be useful if the important information outputted by `gudrun_dcs`, as well as the information outputted in the `.gud` files, was presented after each run.

PR #167, opened 2021-10-18
Show result of gudrun_dcs in GUI.

### LINKED ISSUE #166: Display Gudrun output in GUI?
Rather than relying on the console for the vast text output of `gudrun_dcs`, and since we're already capturing it (?), we could hide the console completely and show it in a separate main tab in the GUI.

jswift-stfc:
Simple PR to show the final level of merged DCS and the relative error with the expected DCS level for each sample, within the GUI.

Edit: closes #166 and #148.

### DIFF
--- a/src/gui/widgets/main_window.py
+++ b/src/gui/widgets/main_window.py
@@ -403,7 +403,7 @@ class GudPyMainWindow(QMainWindow):
                 gf = GudFile(path)
                 dcsLevel = gf.averageLevelMergedDCS
                 self.mainWidget.dcsLabel.setText(
-                    f"Average level of merged DCS: {dcsLevel}"
+                    f"DCS Level: {dcsLevel}"
                 )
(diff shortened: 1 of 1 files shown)


src/gui/widgets/main_window.py
```diff
@@ -366,60 +370,83 @@ def updateCompositions(self):
...
             )
+            if not os.path.exists(path):
+                path = os.path.join(
+                    self.gudrunFile.instrument.GudrunInputFileDir, path
+                )
+            if os.path.exists(path):
+                gf = GudFile(path)
+                dcsLevel = gf.averageLevelMergedDCS
+                self.mainWidget.dcsLabel.setText(
+                    f"Average level of merged DCS: {dcsLevel}"
```
trisyoungs, line 404: Just a small suggestion to reduce text and simplify:
```suggestion
                    f"DCS Level: {dcsLevel}"
```

Issue #168, opened 2021-10-18
Use exponential spin box for composition table


jswift-stfc:
The exponential spin box can be used for the abundance column in the composition table.

PR #169, opened 2021-10-18
Exponential spinbox

jswift-stfc:
This PR implements an `ExponentialSpinBox` for use throughout the GUI. For now, this is only used in place of `QDoubleSpinBoxes` - in future, as mentioned in #168, it can be used as a delegate.


trisyoungs (changes requested):
Looks good, but there's a couple of tweaks to be made in the regexp validator:

1. You can't have zero characters in your spin box - which means you can't delete the last remaining digit!
2. I can enter this: `2.0e4----`
3. And this: `1.0e-+-+`

PR #171, opened 2021-10-19
Chart fixes

jswift-stfc:
Small PR to resolve the following problems:

- Debug output being printed to the console.
- Large space being left between plots and results labels.
- Added combo boxes to be used as stubs for switching between plotting modes in the future.

### DIFF
--- a/src/gui/widgets/main_window.py
+++ b/src/gui/widgets/main_window.py
@@ -1,6 +1,5 @@
 from queue import Queue
 from PySide6.QtCore import QFile
-from PySide6.QtGui import QPainter
 from src.gui.widgets.exponential_spinbox import ExponentialSpinBox
 from src.gudrun_classes.tweak_factor_iterator import TweakFactorIterator
 from src.gudrun_classes.wavelength_subtraction_iterator import (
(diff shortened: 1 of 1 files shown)


src/gui/widgets/main_window.py
```diff
@@ -155,53 +156,67 @@ def initComponents(self):
...
 
-        self.mainWidget.sampleStructureFactorChartView = GudPyChartView(
+        self.mainWidget.sampleTopPlot = GudPyChartView(
             self.mainWidget
         )
 
-        self.mainWidget.sampleStructureFactorChartView.setRenderHint(
+        self.mainWidget.sampleTopPlot.setRenderHint(
             QPainter.Antialiasing
         )
```
trisyoungs, line 165: Since you're always setting this, can you do it in the constructor for `GudPyChartView`?

trisyoungs (approved):
One small comment in the code, and a suggestion here - if you move the plot style combo box to sit above the plots, in principle you could remove the title from the plots as it's basically duplicating the combo box text, and claim a bit of plotting real estate back.

jswift-stfc:
> One small comment in the code, and a suggestion here - if you move the plot style combo box to sit above the plots, in principle you could remove the title from the plots as it's basically duplicating the combo box text, and claim a bit of plotting real estate back.

Yes! All space is valuable, this is a good suggestion.

PR #175, opened 2021-10-19
Reorganise Instrument Page

trisyoungs:
In the interests of actual collaboration by me on the project (and also in part because capturing this stuff in words takes a lot longer) this PR reorganises the widgets on the Instrument page.

Some notes:
1. I had to remove a *lot* of minimum sizes from widgets - this was blocking the layout from flowing properly.
2. Controls are now all (hopefully) sensibly organised from the perspective of an instrument scientist
3. The groups parameters checkable group has been removed in favour of a small tab widget
4. There was no spin box for the logarithmic step size - this has been added.

Important - this only adjusts the UI file!  No attempts to fix code breakage have been made, or implementing missing features.  I leave that to the true Python expert to add on to this PR.

### DIFF
--- a/src/gui/widgets/ui_files/mainWindow.ui
+++ b/src/gui/widgets/ui_files/mainWindow.ui
@@ -212,6 +212,12 @@
                  </item>
                  <item row="10" column="1">
                   <widget class="QComboBox" name="dataFileTypeCombo">
+                   <property name="sizePolicy">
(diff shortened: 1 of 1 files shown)


trisyoungs:
Ah, missed that one! Just pushed a commit to fix that as well as tweak some other labels.

trisyoungs:
I can do better! Last commit.

PR #202, opened 2021-10-28
Track purge_det

### LINKED ISSUE #131: Track progress of purge_det

jswift-stfc:
PR addressing the issue of tracking the progress of `purge_det` using the progress bar. This seems to be working, however could do with some testing to see how accurate/consistent it is.
Closes #131.

### DIFF
--- a/src/gui/widgets/main_window.py
+++ b/src/gui/widgets/main_window.py
@@ -804,16 +804,15 @@ class GudPyMainWindow(QMainWindow):
         stdout = bytes(data).decode("utf8")
 
         path = os.path.join(
-            self.gudrunFile.instrument.startupFileFolder,
+            self.gudrunFile.instrument.GudrunStartFolder,
             self.gudrunFile.instrument.groupFileName
         )
         numGroups = 0
         if os.path.exists(path):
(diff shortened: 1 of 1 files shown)


trisyoungs:
Making progress.  This now tracks the purge through the raw files, but there is a significant period at the beginning where something is happening, but it is not being tracked.  Have to look in to the exact output of purge_det to make a suggestion on what else to capture here...

jswift-stfc:
> Making progress. This now tracks the purge through the raw files, but there is a significant period at the beginning where something is happening, but it is not being tracked. Have to look in to the exact output of purge_det to make a suggestion on what else to capture here...

It looks like it first does something with the specified groups file. I'll add a check for that, and see if it smooths it out a bit.

trisyoungs (approved):
Now get some "preliminary" indication of progress, but there's a big pause after that before the final "splurt" of progress happens.  However, this may just be a delay in windows, as on Linux I don't really see that. 

So, :+1:!

jswift-stfc:
Something to remember when we come back to finalising this feature!

Issue #203, opened 2021-10-29
Broken lambda functions


jswift-stfc:
It seems upgrading to PySide6 has caused some issues with the context menu within the composition table, which creates dynamic `QActions` in a loop. This functionality has been broken by PySide6.

Issue #201, opened 2021-10-27
Clicking a SampleBackground raises a KeyError


jswift-stfc:
Problem as described in the title. It looks like GudPy attempts to plot the SampleBackground, incorrectly thinking that it is a Sample.

PR #204, opened 2021-10-29
Objectify composition

### LINKED ISSUE #118: Objectify definition of composition
Currently, the composition is defined in a "matrix" form - by giving each element, along with its mass number and abundance.
It would be **very** useful to change the way we define compositions to a more user-friendly way - one way could be by entering elements and ratios - and then iterating to correct these ratios (see #117).

jswift-stfc:
This PR implements objectification of composition.
It enables the user to define 'components' which they can use in compositions with ratios. When writing out the `GudrunFile`, it is translated into an exact composition based on the ratio and the component.

Stub for the context menu has been setup, although it is not entirely working due to #203 - which will be addressed in another PR.

Closes #118, #201.


src/gudrun_classes/composition.py
```diff
@@ -1,47 +1,84 @@
...
+    def translate(self):
+        elements = []
+        self.elements = []
+        for component in self.weightedComponents:
+            elements.extend(component.translate())
+        self.sumAndMutate(elements)
+
+    def sumAndMutate(self, elements):
+        for element in elements:
+            exists = False
```
trisyoungs, line 68: Bit of commenting on these two functions would be useful.

trisyoungs (approved):
I like this a lot.  Lots of suggestions to make, but purely in the name of making a good thing better:

1. Add a descriptive text underneath the `Use component/molecule definitions?` to explain briefly what will happen if selected.
2. Change `Use component/molecule definitions?` to `Use component definitions for sample atomic compositions?` - we're not touching the normalisation or the containers at this point.
3. Adding a component in a Sample composition defaults to a ratio of `0.0` - `1.0` would be more useful!
4. A really useful feature to have when setting up components would be to be able to input a chemical formula (e.g. C6H6, H2O, C6O6D3H3` etc.) and have the table populated with those numbers.  This is probably a separate issue, but will make a lot of people happy.
5. Add an option to reduce the generated atomic compositions to a fractional composition - at present the generated atomic compositions are the literal sums of atoms within components multiplied by their ratio, but sometimes it's very useful to see this as a normalised set of fractional compositions which sum to 1.0 (makes it easy to see what percentage of a given atom you have in the sample).

jswift-stfc:
When you say "we're not touching the normalisation or the containers at this point", does that mean that we shouldn't allow these components to be used for the normalisation/containers?

PR #207, opened 2021-11-04
Normalise atomic compositions

### LINKED ISSUE #206: Add option reduce atomic compositions to fractional compositions
Add an option to reduce the generated atomic compositions to a fractional composition - at present the generated atomic compositions are the literal sums of atoms within components multiplied by their ratio, but sometimes it's very useful to see this as a normalised set of fractional compositions which sum to 1.0 (makes it easy to see what percentage of a given atom you have in the sample).

jswift-stfc:
This PR implements a checkbox which enables/disables normalisation of sample atomic compositions ( that have been generated by using user-defined components). When enabled, the abundance of each element is normalised as a ratio of 1. Also adds option to duplicate components.

Closes #206.


trisyoungs (changes requested):
Works nicely overall - a couple of usage "annoyances" that should be addressed:

1. If using components, the "Ratio Composition" tab is always force-shown when switching between Samples - if I'm trying to look at all of the atomic compositions, for example, this is a bit intrusive.
2. When no components are defined it seems like the atomic composition table is not updated, and still shows the values from the previously-selected Sample.
3. If I add a single component to a Sample composition, it defaults to a ratio of 1.0, but the atomic composition still reflects the original data from the file. If I then edit the ratio (even setting it to 1.0 again) the atomic composition is correctly updated.

trisyoungs (changes requested):
Getting there!

I note that you can have multiple components with the same name - this is something we should disallow!

Also, can we have an icon for the `Duplicate` button, rather than text?

trisyoungs (changes requested):
The normalisation stuff all works well.  There still remains the issue of disallowing identical component names from being specified, and I've just noticed a chunky bug!  If I specify two components `H2O` and `D2O` (what else!) and make a mixture of them, the isotope information is lost.  In other words, if I do a 1:1 mix of the two components I get a total atomic composition of `H4O2`, when it should be `H2D2O2`.  Also, we should add a test for the compositions so this type of error can be caught.

jswift-stfc:
> The normalisation stuff all works well. There still remains the issue of disallowing identical component names from being specified, and I've just noticed a chunky bug! If I specify two components `H2O` and `D2O` (what else!) and make a mixture of them, the isotope information is lost. In other words, if I do a 1:1 mix of the two components I get a total atomic composition of `H4O2`, when it should be `H2D2O2`. Also, we should add a test for the compositions so this type of error can be caught.

Which names are you using for components? When I add two components and try and name them both the same, the second one doesn't get applied. Are you referring to when multiple components are added and given the default "Component" name?

jswift-stfc:
> > The normalisation stuff all works well. There still remains the issue of disallowing identical component names from being specified, and I've just noticed a chunky bug! If I specify two components `H2O` and `D2O` (what else!) and make a mixture of them, the isotope information is lost. In other words, if I do a 1:1 mix of the two components I get a total atomic composition of `H4O2`, when it should be `H2D2O2`. Also, we should add a test for the compositions so this type of error can be caught.
> 
> Which names are you using for components? When I add two components and try and name them both the same, the second one doesn't get applied. Are you referring to when multiple components are added and given the default "Component" name?

Regarding the second point, I can't seem to replicate it. Any chance you can try and tell me the exact steps that got you there?

trisyoungs:
> > The normalisation stuff all works well. There still remains the issue of disallowing identical component names from being specified, and I've just noticed a chunky bug! If I specify two components `H2O` and `D2O` (what else!) and make a mixture of them, the isotope information is lost. In other words, if I do a 1:1 mix of the two components I get a total atomic composition of `H4O2`, when it should be `H2D2O2`. Also, we should add a test for the compositions so this type of error can be caught.
> 
> Which names are you using for components? When I add two components and try and name them both the same, the second one doesn't get applied. Are you referring to when multiple components are added and given the default "Component" name?

Ah, well that's cheating!  I was referring to the situation where you add multiple components at once, and they all get the same name.  We should disallow that from the very beginning - just need to add a number to the end of the name string and keep checking until you find one that doesn't exist.

jswift-stfc:
> Ah, well that's cheating! I was referring to the situation where you add multiple components at once, and they all get the same name. We should disallow that from the very beginning - just need to add a number to the end of the name string and keep checking until you find one that doesn't exist.

Aha! easily rectified.

trisyoungs:
> > > The normalisation stuff all works well. There still remains the issue of disallowing identical component names from being specified, and I've just noticed a chunky bug! If I specify two components `H2O` and `D2O` (what else!) and make a mixture of them, the isotope information is lost. In other words, if I do a 1:1 mix of the two components I get a total atomic composition of `H4O2`, when it should be `H2D2O2`. Also, we should add a test for the compositions so this type of error can be caught.
> > 
> > 
> > Which names are you using for components? When I add two components and try and name them both the same, the second one doesn't get applied. Are you referring to when multiple components are added and given the default "Component" name?
> 
> Regarding the second point, I can't seem to replicate it. Any chance you can try and tell me the exact steps that got you there?

OK, here goes:
1. Add a component called "H2O", and set its atomic composition to H(0) = 2, O(0) = 1.
2. Duplicate the "H2O" component, call it "D2O", and change the mass number of H from 0 to 2.
3. Activate "Use component definitions..."
4. Go to a sample, and in the Ratio Composition add "H2O" with a weight of 1.0, and "D2O" with a weight of 1.0.
5. Check the Exact Composition - it states H(0) = 4 and O(0) = 2, whereas it should be H(0) = 2, H(2) = 2, and O(0) = 2.

jswift-stfc:
I think I understand what's going wrong. Sounds like a shared reference problem..

trisyoungs (approved):
Looking good! :+1:

PR #208, opened 2021-11-05
Parse chemical formulae

### LINKED ISSUE #205: Allow components to be defined using chemical formula
It would be very useful to enable components (implemented by #204) to be defined using chemical formula, e.g. H2O, which are then parsed to construct the composition.

jswift-stfc:
This PR implements parsing of chemical formulae as component names.

If a chemical formula is detected as a component name, the user will be asked, in the form of a dialog (with the extracted isotopes/elements), if they wish to use the parsed composition.

Closes #205.


trisyoungs (changes requested):
It's an excellent start - the users are going to love it.  Some routes for improvement:

1. Currently this only works for integer numbers in the composition - putting in a decimal value breaks the parsing (but it still gets detected as a formula).  Fractional compositions will be a common occurrence.
2. Since you're not checking for legitimate chemical elements, I can put in bogus ones and still get it parsed.  As a future enhancement of this feature, we should consider checking elements against a periodic table reference.
3. Lazy chemists (like me!) will want to be able to specify deuterated chemicals in the same way, which means we need to be able to detect the special case of `D`, and interpret it as `H` with a mass number of `2`.

src/gudrun_classes/composition.py
```diff
@@ -1,15 +1,74 @@
...
+
+    def parseElement(self):
+        symbol = self.parseSymbol()
+        abundance = self.parseAbundance()
+        if symbol == "D":
+            symbol = "H"
+            abundance = 2.0
+        from src.gudrun_classes import config
+        if symbol and abundance and symbol in config.massData.keys():
+            return Element(symbol, 0, abundance)
```
trisyoungs, line 37: For `D`, this should change the mass number (the `0` that you pass in `return Element(...`), not the `abundance`

src/gui/widgets/dialogs/composition_dialog.py
```diff
@@ -0,0 +1,60 @@
...
+                    current_dir, "..", "ui_files", "compositionDialog.ui"
+                )
+            )
+        loader = QUiLoader()
+        self.widget = loader.load(uifile)
+        self.widget.setWindowTitle("Compositon")
+
+        self.widget.infoLabel.setText(
+            f"{self.component.name} looks like a chemical formula.\n"
+            f" Do you want it's isotopes resolved automatically?"
```
trisyoungs, line 54:
```suggestion
            f"Do you want the atomic composition resolved automatically?"
```

src/gui/widgets/dialogs/composition_dialog.py
```diff
@@ -0,0 +1,60 @@
...
+            f" It looks like:\n"
```
trisyoungs, line 55:
```suggestion
            f"It looks like:\n"
```

tests/test_chemical_parser.py
```diff
@@ -0,0 +1,114 @@
...
+    def testD2OParse(self):
+
+        formula = "D2O"
+        component = Component(formula)
+        component.parse()
+
+        self.assertEqual(len(component.elements), 2)
+
+        self.assertEqual(component.elements[0].atomicSymbol, "H")
+        self.assertEqual(component.elements[0].massNo, 0)
```
trisyoungs, line 40: Once earlier bug is fixed, this will be:
```suggestion
        self.assertEqual(component.elements[0].massNo, 2)
```

tests/test_chemical_parser.py
```diff
@@ -0,0 +1,114 @@
...
+        self.assertEqual(component.elements[1].massNo, 0)
+        self.assertEqual(component.elements[1].abundance, 1.0)
+
+        self.assertEqual(component.elements[2].atomicSymbol, "Ar")
+        self.assertEqual(component.elements[2].massNo, 0)
+        self.assertEqual(component.elements[2].abundance, 33.0)
+
+        self.assertEqual(component.elements[3].atomicSymbol, "Au")
+        self.assertEqual(component.elements[3].massNo, 0)
+        self.assertEqual(component.elements[3].abundance, 26.5)
```
trisyoungs, line 105: Lots of "positive" tests here, but it would be worth checking the other end of the spectrum - i.e. what happens when the chemical formula contains an element that isn't listed in the periodic table data?

trisyoungs (approved):
Couple of points to address before merging, but otherwise good.

trisyoungs:
Just noticed a minor annoyance with this! Once a chemical formula has been parsed, the component list resets and the first item gets re-selected, rather than remaining on the one that I've just changed the name of....

PR #216, opened 2021-11-24
Generic fixes

### LINKED ISSUE #215: Plotting when running a single sample
When a single sample is ran and eventually plotted, the RDF plot does not work.

jswift-stfc:
Various bug fixes and minor enhancements.
Notable enhancements are that the suggested tweak factor is now shown in the GUI, and instrument configurations have been added which can be loaded and used (although for now, NIMROD is the only stable one).

Closes #215.

### DIFF
--- a/.azure-pipelines/templates/test.yml
+++ b/.azure-pipelines/templates/test.yml
@@ -36,7 +36,7 @@ steps:
   - bash: |
       curl -L https://github.com/disorderedmaterials/Gudrun/releases/download/${{ parameters.gudrunTag }}/binaries-${{ parameters.gudrunTag }}-${{ parameters.osName }}.zip > binaries.zip
       unzip binaries.zip
-      mv binaries-${{ parameters.gudrunTag }}-${{ parameters.osName }} bin
+      mv binaries-${{ parameters.gudrunTag }}-${{ parameters.osName }}/* bin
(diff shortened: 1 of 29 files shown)


trisyoungs:
A crash!

```
  File "main.py", line 5, in <module>
  File "src\gui\gudpy.py", line 23, in main
  File "src\gui\gudpy.py", line 10, in __init__
  File "src\gui\gudpy.py", line 16, in initComponents
  File "src\gui\widgets\main_window.py", line 113, in __init__
  File "src\gui\widgets\main_window.py", line 315, in initComponents
AttributeError: 'PySide6.QtWidgets.QMainWindow' object has no attribute 'new_'
[11800] Failed to execute script 'main' due to unhandled exception!
```

trisyoungs (changes requested):
Lots of nice new stuff in here.  I noticed a couple of small things:

1. I think that locating the "Suggested Tweak Factor" label and value underneath the DCS / level percentage would look better - at the minute the variability in the tweak factor value keeps resizing the whole right-hand panel.
2. Choosing the "New" option doesn't clear any data displayed on the "Plots" tab.

jswift-stfc:
Probably makes more sense that way. I shall fix the second point. A non-related question, but I think you said the other day that you would expect the output files to be produced in the directory of the original input file, is that right? Currently they are produced where GudPy is being run.

trisyoungs:
I would expect that only based on accepted behaviour of the original GudrunGUI.  It also feels sensible to have the output generated in the same directory as the input!

jswift-stfc:
It does make more sense. I'll add that to this PR.

PR #238, opened 2021-11-29
Plot expected DCS level

### LINKED ISSUE #226: Structure factor plot should show expected DCS level
Plots of structure factors should show the expected DCS level (i.e. that calculated from the composition, density, and tweak factor) rather than the actual limit derived from the processed data.

jswift-stfc:
Small PR to plot the expected DCS level, closes #226.

### DIFF
--- a/src/gui/widgets/gudpy_charts.py
+++ b/src/gui/widgets/gudpy_charts.py
@@ -404,6 +404,7 @@ class GudPyChart(QChart):
                 )
                 pen = QPen(dcsSeries.pen())
                 pen.setStyle(Qt.PenStyle.DashLine)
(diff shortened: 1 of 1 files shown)


trisyoungs:
While we're here, there are two other small improvements to make before merging:
1. The dashed line representing the expected DCS level is really hard to see - can we make it a bit thicker?
2. In the legend we should refer to the "Expected level" rather than "DCS level"

jswift-stfc:
Can definitely make it thicker. I got a bit ahead of myself and forgot to rename that!! Apologies. I'll make it thicker and sort that out, then merge this in.

PR #241, opened 2021-11-29
Units and minor GUI fixes/enhancements

### LINKED ISSUE #218: Clarify horizontal / vertical controls
Certain values - e.g. the incident beam edges and scattered beam edges - are currently represented by a group of four double spin boxes arranged horizontally (as in the old GudrunGUI) which makes it unclear which controls represent the left, top, right, and bottom edges.

These controls should be layed out such that their intention is more clear visually.

### LINKED ISSUE #222: Dialog with gudrun_dcs output
Since we already capture the output from `gudrun_dcs`, we should add a menu option to display it in a QTextEdit.

### LINKED ISSUE #228: Density unit names
Naming of enumerations for the density units are not quite correct - e.g. `ATOMIC` should really be `atoms / A3` (where the `A` is really an Angstrom).  This might cause problems with using the raw enum names as the text in the combo boxes, so an alternative way of linking the name to the enum might need to be sought?

### LINKED ISSUE #229: Enumerate Top Hat Width options
The top hat width in the Fourier transform may take a negative number to represent a slightly different approach.  We should enumerate this to make things clear to the user what it is they are requesting.

### LINKED ISSUE #231: Show total sample thickness
Add a simple label underneath the upstream / downstream thicknesses of the Sample to calculate and show the total thickness.

### LINKED ISSUE #232: Units for upstream / downstream thickness
No units currently shown on upstream / downstream thickness controls?

### LINKED ISSUE #237: Modify PurgeFile class
The PurgeFile class does all it's "class setup" inside the `__str__` method, however if we wanted to access some of it's attributed (e.g. it's data files), the` __str__` method must first be called. This is obviously not good practice. The setup code should be moved to a separate method, for instance `collectDataFiles` to be called within the constructor. Also, it might be good here to utilise the `GudPyFileLibrary` class, as this should collect all the data file needed for the purge.

jswift-stfc:
This PR implements the following:

- Units for geometric attributes (thickness etc.). Closes #232.
- Calculates and shows total thickness. Closes #231.
- Fix bugs with `PurgeFile`, which previously broke the progress bar. Closes #237.
- Shows the correct units in the combo boxes for density units. Closes #228.
- Enumeration added for `FTModes`, which is used to determine when to negate etc. the `Top-Hat Width`. The name of the enumeration/corresponding member attribute of the `Sample` class may not be an accurate representation, however. Closes #229.
- View output of `purge_det`/`gudrun_dcs` in a dialog, for debugging. Closes #222.
- Also closes #218.


trisyoungs (changes requested):
Lots of good fixes in here, and lots of small comments to follow! :)

1. For the units, I wonder if we're better off making use of the `suffix` property of the spin boxes?  Having the units in group box titles or labels feels a little separated.
2. For the total thickness readout, why not just use a simple `QLabel` rather than a disabled spin box?
3. As discussed, the Angstrom symbol in the density units combo looks horrible! Not your fault, but worth trying to rectify.

jswift-stfc:
Seems like the Angstrom symbol issue comes with the Windows-specific fonts. I think maybe we should choose a different font for Windows, rather than the defaults. This is what `Tahoma` looks like:
image

Would it be worth using the same font across all platforms? Or should we keep the default fonts on the other platforms, to feel more native?

trisyoungs:
IMO, I think we have to go native for the best experience all round.

trisyoungs (commented):
I have to be honest - I think the custom font makes it look odd, especially given the titlebar, menus, and file selection dialogs all use the native Windows fonts. Also, the Angstrom symbol still looks weird (too thin)!

We tried, but I think that the font change should be dropped. Once done, this can be merged in.

jswift-stfc:
> I have to be honest - I think the custom font makes it look odd, especially given the titlebar, menus, and file selection dialogs all use the native Windows fonts. Also, the Angstrom symbol still looks weird (too thin)!
> 
> We tried, but I think that the font change should be dropped. Once done, this can be merged in.

Maybe something we can revisit at a later point! For now I'll remove the font and merge it in.

PR #244, opened 2021-12-01
Export archive

### LINKED ISSUE #235: Export archive file containing mint01 data
An option to save a zip archive of, principally, the mint01 files would be a useful feature.  As part of this, having the option to rename the mint-1 files after the SAMPLE names (rather than the raw file names) would be helpful.

jswift-stfc:
This PR implements exporting a data archive of mint01 files relevant to the current input file. Also allows the exported files to be renamed to the sample name. Closes #235.


trisyoungs (changes requested):
Nice feature!  Some minor suggestions:

1. The `File` menu item just says `Export`.  Export what? Change this to `Export mint01 Archive`
2. The `Export As` option in the Export dialog should really enforce the `zip` extension, and have that as the file type the file selector recognises.
3. The `File` menu itself is getting a bit disorganised.  I'd suggest separating groups of related menu items with `QSeparator`s.

trisyoungs:
And a bug I've just spotted! If I cancel the file dialog in the `Export As` workflow, GudPy still goes ahead and saves a zipfile named after the input file.

trisyoungs (changes requested):
A new bug!  If I have "Rename files to sample names" checked, I don't get a confirmation message dialog when the zip is exported (either with `Export` or `Export As`).

trisyoungs:
```
Traceback (most recent call last):
  File "src\gui\widgets\dialogs\export_dialog.py", line 122, in export
  File "src\gui\widgets\dialogs\export_dialog.py", line 105, in performExport
  File "src\gudrun_classes\file_library.py", line 158, in exportMintData
FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'C:\\Users\\axr61803\\Desktop\\NIMROD-water\\NIMROD00016608_H2O_in_N9.mint01' -> 'H2O_Can_N9.mint01'
```

jswift-stfc:
I think maybe the solution to this is to rename the data file inside the archive, rather than the initial file? Would that be okay?

trisyoungs:
I guess so.  That sounds neater.

PR #246, opened 2021-12-06
Parse Isotopes

### LINKED ISSUE #210: Allow mass number to be specified in chemical formulae
A useful feature to enhance the chemical formula parser (#208) would be for it to be able to identify isotopes/defined mass numbers. This could be done by allowing a mass number of an element to be defined within square brackets.

For instance, the following: `H[2]2O`, would be resolved as H(2) 2.0, O(0) 1.0.

### LINKED ISSUE #245: Calculate Expected DCS Level
The expected DCS level can be calculated using the composition. This can be calculated by `scat/4.0/pi`, where scat is the average bound scattering cross section.

jswift-stfc:
This PR implements parsing of isotopes using chemical formulae when defining components. Closes #210.
Also implements calculation of the expected DCS level, without having to run Gudrun and check the output of the .gud files, closes #245.


trisyoungs (changes requested):
Looks good. One small UX issue - if I type in a formula with an invalid mass number for an isotope, I don't get an error message.  GudPy just quietly fails to parse the component name.  It might be nicer to at least show a warning about unrecognised mass numbers for elements, and inform the user they've made a mistake.

Expected DCS level label is marvellous! I feel like we need to reorganise that corner a bit now, but we can save that for a future tweak.

jswift-stfc:
> 

Sounds reasonable, I'll add a small warning informing of the error in the formula.

Issue #227, opened 2021-11-27
Structure factor plotting options


trisyoungs:
Currently we have the option to plot mint01 and mdcs01 files together.  We should split this into separate options for mint01 only, mdcs01 only, or mint01/mdcs01 together.  This option should be available for both individual samples as well as the combined plotting panel.

jswift-stfc:
Should these options look something like this?

`Structure Factor (mint01, mdcs01)`
`Structure Factor (mint01)`
`Structure Factor (mdcs01)`

trisyoungs:
Let's be slightly more explicit and say:
```
Structure Factor (mint01 & mdcs01)
Structure Factor (mint01 only)
Structure Factor (mdcs01 only)
```

PR #255, opened 2021-12-10
Run containers as samples

### LINKED ISSUE #220: Run containers as samples
A common workflow in the original GudrunGUI is to also add all containers as samples in order to check levels / alignment etc. In the present GUI we could simply add a menu option to do a one-shot run of all containers as if they were samples, copying their composition and run files and using sensible defaults for everything else. This leads us to having plotting capability side-by-side on the Container page.

An alternative or additional option would be to have a button on the Container page itself giving the option to "Run as Sample".

jswift-stfc:
This PR implements running containers as samples, either all at once using a menu option, or individually using check boxes on each container page. Also offers functionality to compare a container to a sample through context menus. Closes #220.


trisyoungs (changes requested):
This is a nice addition, but I think the UI could do with a little bit of tweaking.

1. The `Results` group is always visible on the container page, even if we haven't run it as a sample. I think it should be put into a separator so it can be resized a la the Sample tab.
2. Currently we don't get a g(r) or d(r) because, I assume, the Top Hat width isn't set to a non-zero value. You may have to duplicate the Fourier transform controls from the Sample page, but perhaps make the `Run as Sample` a checkable group box, and put the Fourier transform controls in that group so that they can be disabled when the group box isn't checked.
3. An additional option for the main `Plots` tab would be to show all the structure factors from the cans (if they've been run, of course).

jswift-stfc:
1. Should I maybe set the size of the `Results` group to 0, within the splitter, by default? If there are plots available, then set it to another default value, ensuring visibility?
2. I was disabling Fourier Transform by default for the containers being ran as samples. I like the idea of having a group box for the Fourier Transform controls. How should I default them values? Should I copy them from the associated sample maybe?
3. That makes sense. Would this be also useful for `RDF`?

trisyoungs:
1. Should work, although I've had issues with splitters and doing this in the past.
2. Ah, but when do you copy them from the sample? As long as the controls are there, that should be fine. Let the user worry about it.
3. Yeah, why not!

jswift-stfc:
For now I added standard Structure Factor / RDF plots, i.e. Structure Factor shows mint01/mdcs01/dcs level all in one plot. Should I add the enhanced plotting options, as suggested in #227 for samples, for the containers?

trisyoungs (approved):
:+1:

PR #260, opened 2021-12-13
Track parameters

jswift-stfc:
PR implementing basic tracking of parameters of samples used whilst running.

- .sample files are written out for each sample whilst running, to the input file directory.
- .sample files can be loaded as samples into the GUI.
- parameter files can be exported within the mint01 archive.


trisyoungs (commented):
This doesn't appear to write any `.sample` files!  Is your main `write_out` function ever called with `writeParameters=True`?

jswift-stfc:
I should pay more attention when resolving merge conflicts.. especially at 8AM on a Monday morning!! I accidentally removed the argument that should be passed to write out the parameters.. Fixed now.

trisyoungs (changes requested):
OK, this looks good.  I think we should add more information to the `.sample` files, though. Ultimately, this information could form part of the associated data for a publication, and the sample parameters on their own are not enough to reproduce the processing with Gudrun.  I think we'll have to take the plunge and write out a full input file with the single Sample in.

jswift-stfc:
That would include the associated containers, right?

PR #267, opened 2021-12-13
Beam edges

### LINKED ISSUE #266: Beam profile controls
Just noticed that:

1. The beam profile values for left and bottom can't go negative as the exponential spin boxes are limited to 0.0.
2. The top and bottom controls are laid out, well, bottom-to-top!  The bottom (probably negative) value should be at the bottom, but it currently looks like its at the top.
3. Need tool tips on the controls to explicitly indicate which edge is represented.

jswift-stfc:
Small PR which closes #266. 

- Fixed bug with organisation of edges into the correct categories (left/right/bottom/top).
- Removed hard limit of 0 for the exponential spin boxes for the left/bottom edges.
- Tool tips added to make it clearer which spin box corresponds to which edge.


trisyoungs (approved):
:+1:

PR #268, opened 2021-12-14
Refactor gud file parser

### LINKED ISSUE #254: Refactor GudFile parser
The `GudFile` parser uses line numbers for parsing, similar to how the `GudrunFile` parser previously did. See #70. This seems a little fragile, and for the first time ever, today I experienced a crash caused by this parser.

jswift-stfc:
The `GudFile` parser has been refactored to parse in a similar fashion to the `GudrunFile` parser. Closes #254.

### DIFF
--- a/src/gudrun_classes/gud_file.py
+++ b/src/gudrun_classes/gud_file.py
@@ -189,98 +189,108 @@ class GudFile:
 
         # Simple cases, we can just extract the stripped lines.
 
-        self.name = self.getNextToken().strip()
-        self.consumeTokens(1)
+        try:
 
-        self.title = self.getNextToken().strip()
-        self.consumeTokens(1)
-
-        self.author = self.getNextToken().strip()
-        self.consumeTokens(1)
-
-        self.stamp = self.getNextToken().strip()
-        self.consumeTokens(1)
-
-        self.atomicDensity = float(
-            self.getNextToken().split()[-1].strip()
-        )
-
-        self.chemicalDensity = float(
-            self.getNextToken().split()[-1].strip()
-        )
-
-        self.averageScatteringLength = float(
-            self.getNextToken().split()[-1].strip()
-        )
-
-        self.averageScatteringLengthSquared = float(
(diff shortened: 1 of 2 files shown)


src/gudrun_classes/gud_file.py
```diff
@@ -143,78 +184,115 @@ def parse(self):
...
-        i = 1
-        while not line.isspace():
-            self.groups.append(line)
-            line = self.contents[19 + i]
-            i += 1
+        try:
 
-        self.groupsTable = "".join(self.groups)
+            self.name = self.getNextToken().strip()
+            self.consumeTokens(1)
```
trisyoungs, line 195: I want to say here that when you `get` a token/line you are implicitly comsuming it, which would reduce a lot of these to one-line calls. That said, it probably breaks your logic in other places!

src/gudrun_classes/gud_file.py
```diff
@@ -128,6 +128,47 @@ def __init__(self, path):
         # Parse the GudFile
         self.parse()
 
+    def getNextToken(self):
```
trisyoungs, line 131: In all these new functions you refer to "tokens", when really you are handling entire lines at a time.  Semantically, I immediately think of a "token" as a "bit of a line" (an "argument", if you will).

trisyoungs (commented):
Beautifully clean, but perhaps naming needs to be addressed? See comments below.

src/gudrun_classes/gud_file.py
jswift-stfc: The consume here refers to where the line after the line extracted is empty. Maybe it would be better to simply ignore the empty lines when initally reading in the contents of the file?

trisyoungs: Ah. Well, if you can absolutely guarantee that an empty line is not important in the Gudrun input file, I guess you can automatically parse through them (or, to get the best of both worlds, put an optional `bool` argument on your `getNextToken` function defining whether to ignore blank lines or not.

PR #269, opened 2021-12-14
Modification status fix

### LINKED ISSUE #265: File modified indicator
The behaviour of the file changed indicator (`*` after the filename in the title bar) seems to be a little bit inconsistent.

Principally:
- The `Save` menu item isn't disabled when no modifications to the file have been made.
- Actually editing values on any tab doesn't reliably (if at all) change the modification status (at least not visibly on the title bar)

jswift-stfc:
This PR attempts to fix the modification status functionality. Rather than setting the window title after a modification, we now use the `setWindowModified` method of the main window, to apply the modification status and have it persist in the title. Closes #265.

### DIFF
--- a/src/gui/widgets/main_window.py
+++ b/src/gui/widgets/main_window.py
@@ -410,7 +410,7 @@ class GudPyMainWindow(QMainWindow):
                     del self.gudrunFile
                 self.gudrunFile = GudrunFile(path=filename)
                 self.updateWidgets()
-                self.mainWidget.setWindowTitle(self.gudrunFile.path)
+                self.mainWidget.setWindowTitle(self.gudrunFile.path + "[*]")
             except ParserException as e:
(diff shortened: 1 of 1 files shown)


src/gui/widgets/main_window.py
```diff
@@ -879,11 +879,11 @@ def showPreviousOutput_(self):
     def setModified(self):
         if not self.modified:
             if self.gudrunFile.path:
-                self.mainWidget.setWindowTitle(self.gudrunFile.path + " *")
+                self.mainWidget.setWindowModified(True)
```
trisyoungs, line 882: Would be nice to accordingly disable/enable the `Save` menu item in these functions too.

trisyoungs (approved):
Ah, much better! One minor change requested, then it can go.

src/gui/widgets/main_window.py
jswift-stfc: Ah I forgot about that!

PR #270, opened 2021-12-15
Purge with default parameters

### LINKED ISSUE #263: Purging with default parameters
As a user, when I run Gudrun and haven't performed a purge, we ask what should be done.  If I click on "Run with default parameters", I kinda don't want the processing to stop and inform me of the number of detectors making it through the purge. Rather, I want the Gudrunning to happen!

To cater for the case where this "blind" running of the purge might introduce errors due to bad numbers of detectors, we could elect to only raise a dialog *if* the number of detectors comes outside of our acceptable (per-instrument) boundaries, once that functionality is implemented.

jswift-stfc:
This PR updates the way the results of `purge_det` are indicated. Now, a warning is produced, currently for `NIMROD` only, if the number of detectors that made it through the purge is below an acceptable threshold. Also, the number of good detectors is indicated on the `Instrument` page. This closes #263.

### DIFF
--- a/bin/configs/NIMROD.txt
+++ b/bin/configs/NIMROD.txt
@@ -34,5 +34,6 @@ StartupFiles/NIMROD/sears91_gudrun.dat          Neutron scattering parameters fi
 1          Hard group edges?
 0          Number of iterations
 0          Tweak the tweak factor(s)?
+9999        Good detector threshold
 
 }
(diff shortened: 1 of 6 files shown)


trisyoungs (commented):
I am of the opinion that this should be implemented as part of the instrument configuration file work, since it is not necessarily the case that all configurations for a given instrument utilise the same number of detectors in the array.

jswift-stfc:
Makes sense to utilise our instrument configurations that way. Maybe we should have an attribute in the `Instrument` class to specify the minimum acceptable number of good detectors, which is only set when an instrument configuration is loaded?

trisyoungs:
That would work!

PR #274, opened 2021-12-24
Build Singularity image

### LINKED ISSUE #273: Singularity Image
For distributing to Linux, we should build a singularity image, to be compatible with more Linux platforms. Since currently, we include glibc from Ubuntu 20.04, which is not compatible with older versions of Linux.

jswift-stfc:
This PR builds a .sif (Singularity Image File), which can be run using `singularity run`. This image uses Ubuntu 20.04, and aims to resolve dependancy problems with running GudPy across different versions of ubuntu, and flavours of linux. However, the final artifact has not yet been tested.

Closes #273.

### DIFF
--- a/.azure-pipelines/templates/build.yml
+++ b/.azure-pipelines/templates/build.yml
@@ -85,6 +85,5 @@ steps:
   - bash: |
       VERSION=$(grep "VERSION =" gudpy.spec | sed "s/.*\"\(.*\)\"/\1/g")
       sudo singularity build dist/GudPy-$VERSION.sif ci/singularity/ubuntu20.04.def
-      singularity run dist/GudPy-$VERSION.sif
     condition: eq(variables.osName, 'linux')
     displayName: "Build singularity image for linux."
--- a/.azure-pipelines/templates/build.yml
+++ b/.azure-pipelines/templates/build.yml
@@ -34,12 +34,13 @@ steps:
     displayName: "Install OpenGL for linux"
 
   - bash: |
+      rm -rf bin
       curl -L https://github.com/disorderedmaterials/Gudrun/releases/download/${{ parameters.gudrunTag }}/binaries-${{ parameters.gudrunTag }}-${{ parameters.osName }}.zip > binaries.zip
       unzip binaries.zip
       mv binaries-${{ parameters.gudrunTag }}-${{ parameters.osName }} bin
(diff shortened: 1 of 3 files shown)


jswift-stfc:
As of 915e62f390e1a6b3b9e15c38515cf7b228cb5fdf, the sif produced is working as expected. This was achieved by downgrading to singularity version 2.5.2, and making the container writable. As a consequence of this, however, the size of the produced sif has skyrocketed to just under `2GB`.. Going to see if it's possible to reduce this.

trisyoungs (commented):
I'm afraid this still doesn't quite work correctly for me - I can run Gudrun, but it stops prematurely:
Screenshot from 2022-01-12 09-31-57

I don't think that 2Gb is an acceptable size for our users to download. It would be OK for IDAaaS, but for the poor university student at home who wants to run it on their laptop it's a bit much.

jswift-stfc:
> I'm afraid this still doesn't quite work correctly for me - I can run Gudrun, but it stops prematurely: Screenshot from 2022-01-12 09-31-57
> 
> I don't think that 2Gb is an acceptable size for our users to download. It would be OK for IDAaaS, but for the poor university student at home who wants to run it on their laptop it's a bit much.

That's strange that it doesn't work for you, I pulled the artifact and it worked on my linux partition. The premature stopping will be due to lack of permissions to write to `/opt/GudPy/bin/StartupFiles` when Gudrun tries to write the group calibration file. I agree 2GB is way too big. I guess the next idea would be to mount the entire GudPy as a persistent overlay? Or I wonder if it would be enough to do that with the StartupFiles directory, and symlink it to `/opt/GudPy/bin/StartupFiles`?

trisyoungs:
I think just an overlay for StartupFiles will be enough. That's the important one.

jswift-stfc:
> > I'm afraid this still doesn't quite work correctly for me - I can run Gudrun, but it stops prematurely: Screenshot from 2022-01-12 09-31-57
> > I don't think that 2Gb is an acceptable size for our users to download. It would be OK for IDAaaS, but for the poor university student at home who wants to run it on their laptop it's a bit much.
> 
> That's strange that it doesn't work for you, I pulled the artifact and it worked on my linux partition. The premature stopping will be due to lack of permissions to write to `/opt/GudPy/bin/StartupFiles` when Gudrun tries to write the group calibration file. I agree 2GB is way too big. I guess the next idea would be to mount the entire GudPy as a persistent overlay? Or I wonder if it would be enough to do that with the StartupFiles directory, and symlink it to `/opt/GudPy/bin/StartupFiles`?

Also noticed you are using the "View Previous Output" option :D

PR #282, opened 2022-01-17
Store components

jswift-stfc:
PR to implement saving of components to the input file. Also allows parsing of components on loading.


trisyoungs (changes requested):
Looks good, but one minor bug - adding an empty component (i.e. with no atomic composition) saves fine, but can't be reloaded.

trisyoungs (approved):

PR #286, opened 2022-01-21
Gudrun GUI Compatibility

### LINKED ISSUE #284: Gudrun Java GUI Compatibility
The input file produced by GudPy currently is not very compatibile with the Java Gudrun GUI. They should be backwards-compatible.

jswift-stfc:
PR to create a sense of backwards compatibility between GudPy and the Gudrun Java GUI. Files in the GudPy input format can be loaded into the Gudrun GUI.

Saving of components has been modified slightly, where curly braces were previously used, parenthesis are now used - since curly braces break the Gudrun GUI - as these are used as section delimiters there.

Closes #284.


jswift-stfc:
> The EOL issue is now resolved, but I still get a number conversion error when reading in the saved file to Gudrun_GUI. While the conversion error cites the msubw01 filename as the problematic argument, the actual cause of this appears to be the exponential parameter definitions a few lines earlier. This has been discussed before, but while GudPy only writes out 2 values, Gudrun_GUI expects three (even though the last parameter is unused).

Strange.. I can't replicate on OSX. For me, the three values are written out for exponential values. I'm going to try it on Windows now.

trisyoungs:
This was the sif on Linux, for the record.

jswift-stfc:
Did you load the input file into GudPy, alter some value and save it again, and then load it into Gudrun GUI?

trisyoungs:
Exactly that.

PR #296, opened 2022-01-26
Reorganise output

jswift-stfc:
PR to re-organise output (thus far, only gudrun_dcs). Output is now shown in a QPlainTextEdit, which can be navigated through using a combo box. The output of each sample is highlighted when it is in focus.


trisyoungs (changes requested):
Works well enough, but is it a bit clunky?  One thing is for sure - reading black text with a blue highlight on a white background is not a pleasant experience!

I'd like to suggest two alternative layouts:

1. Have a `QListWidget` on the left containing the "sections" and have a `QTextView` on the right which displays just the output from that section (i.e. no highlighting required).
2. Do a similar thing, but have a set of tabs, one for each output section.

Thoughts?  I think I prefer 1) as an option.

jswift-stfc:
I think 1 also. I'll implement that now!

trisyoungs (approved):
Looking good. Just two minor things:

1) I really think the default tab should be "Gudrun Parameters" rather than "Output" - at the beginning there are parameters, but no output!
2) On Windows the output has an extra blank line after each "proper" line of output. If you're adding the lines one by one to the `QTextEdit` I think the default behaviour is for Qt to add on newlines automatically.

PR #297, opened 2022-01-27
Chart updates #1

### LINKED ISSUE #290: Don't show Structure Factor plot with mint01, mdcs01 and DCS level
The Structure Factor plot with the mint01 and mdcs01 files, as well as the expected DCS level, is rather cluttered, when it is used on the "Plot Results" tab, for plotting all samples. I think we should remove this option (but keep it for individual samples), and default to plotting mint01.

### LINKED ISSUE #292: Q-Space / R-Space plots in splitter
Rather than defaulting to showing the Q-Space and R-Space plots in a vertical splitter - default to using a single plot for showing structure factor / RDF etc., with options such as "mint01, mdor01", which then uses the splitter.

### LINKED ISSUE #293: Plot Coordinates
Show the coordinates of the cursor in the plot, in a label, perhaps in the bottom left hand corner of the chart view, or at the cursor.

### LINKED ISSUE #295: PlotMode incorrectly defaulting after running
If a plot mode is selected (other than the default), and then gudrun_dcs is run, the plot mode in the combo box will remain the same, but the default plot will be plotted.

jswift-stfc:
PR to fix up some charting issues and inconveniences.

- No longer plot mint01 and mdcs01 files together, with the expected DCS level on the summary page. Only do this for individual samples/containers. Closes #290.
- Fix bug with incorrect plot mode being shown after running gudrun_dcs - where the plot mode displayed doesn't match the combo box. Closes #295.
- Optionally split Structure Factor / RDF, rather than by default. Closes #292.
- Also, show co-ordinates in the chart of the mouse. Closes #293.

### DIFF
--- a/src/gui/widgets/charts/chartview.py
+++ b/src/gui/widgets/charts/chartview.py
@@ -1,9 +1,9 @@
-from PySide6.QtCharts import QChartView
-from PySide6.QtCore import QRectF, QRect, Qt
+from PySide6.QtCharts import QChartView, QChart
+from PySide6.QtCore import QRectF, QRect, Qt, QPoint
 from PySide6.QtGui import (
     QAction, QClipboard, QCursor, QPainter, QMouseEvent
 )
-from PySide6.QtWidgets import QApplication, QMenu, QSizePolicy
+from PySide6.QtWidgets import QApplication, QMenu, QSizePolicy, QGraphicsTextItem
 from src.gudrun_classes.container import Container
 from src.gudrun_classes.sample import Sample
 from src.gui.widgets.charts.enums import PlotModes, SeriesTypes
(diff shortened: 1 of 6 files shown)


src/gui/widgets/charts/chartview.py
```diff
@@ -93,18 +95,29 @@ def wheelEvent(self, event):
...
-            self.chart().zoom(1 + 0.00000001)
-            self.chart().scroll(-offset.x(), offset.y())
-
-            self.previousPos = event.pos()
-            event.accept()
-        return super().mouseMoveEvent(event)
+                if self.chart().plotArea().contains(event.pos()):
+                    pos = self.chart().mapToValue(event.pos())
+                    self.chart().label.setPlainText(
+                        f"x={round(pos.x(), 2)},y={round(pos.y(), 2)}"
```
trisyoungs, line 114:
```suggestion
                        f"x={round(pos.x(), 4)}, y={round(pos.y(), 4)}"
```

trisyoungs (approved):
Apart from one space and some more accuracy in the displayed coordinate values, good to go.

Issue #211, opened 2021-11-10
Instrument configurations


jswift-stfc:
Creating a new input file from scratch would be tedious - hence, it would be good to have some 'templates' in the form of Instrument Configurations, which a new input file can be created from / can be applied to a newly created input file.

jswift-stfc:
Proper configurations need to be defined for the following instruments:
`POLARIS`
`SANDALS`
`GEM`
`D4C`
`HIPD`
`NPDF`

Currently, a proper configuration has only been set up for `NIMROD` - see #216.

jswift-stfc:
Small update on this matter. `SANDALS` now has it's own configuration included in GudPy.

The current `NIMROD` configuration is for 84 modules, from 2012.
The newly added `SANDALS` configuration is from 2011.

jswift-stfc:
I'm closing this for now, because the functionality exists for configurations to be added easily.

PR #300, opened 2022-02-03
Standard containers

### LINKED ISSUE #288: Standard Containers
Similar to the Instrument Configurations (#211), we can have default Containers which can be selected from when inserting a Container. This will speed up workflows, and prevent parameters/attributes not getting set/not getting set to reasonable values. This stems from #287.

jswift-stfc:
This PR implements a variety of Container templates (taken from the Gudrun StartupFiles), which can be used when inserting a Container, similar to #211. This can be done via the context menu, or the menu bar.

This PR also fixes some issues discovered, whilst implementing the set-out functionality, with the menu bar in the main window and the context menu in the tree view.

Closes #288.


trisyoungs (changes requested):
Nice new feature.  My only comments concern formatting and how we present the available data.

First, the leading path to the container file can be removed - no need to display that (it's ugly!).  Then I think we should consider a hierarchy of folders within the new `containers/` dir to help sort things out.  My initial thoughts for categories would be:

`Sample Changer` - contains the TiZr and Hellma flat plate cells
`Cylindrical` - contains all of the various cylindrical vanadium cylinders.
`Pressure Cells` - a new category to be added at a later date, and containing all of our flat-plate pressure cells (which don't go on the sample changer).

Also, as discussed the different versions of the TiZr flat plate cells (1, 2, and 4 mm) can be captured in a single configuration since the parameters are identical.

jswift-stfc:
> Nice new feature. My only comments concern formatting and how we present the available data.
> 
> First, the leading path to the container file can be removed - no need to display that (it's ugly!). Then I think we should consider a hierarchy of folders within the new `containers/` dir to help sort things out. My initial thoughts for categories would be:
> 
> `Sample Changer` - contains the TiZr and Hellma flat plate cells `Cylindrical` - contains all of the various cylindrical vanadium cylinders. `Pressure Cells` - a new category to be added at a later date, and containing all of our flat-plate pressure cells (which don't go on the sample changer).
> 
> Also, as discussed the different versions of the TiZr flat plate cells (1, 2, and 4 mm) can be captured in a single configuration since the parameters are identical.

Oops, I didn't even notice that the absolute path was being shown.. Regarding merging into a single configuration, should this be reflected in the GUI - or should the three separate options still be offered (for some reason)?

PR #306, opened 2022-02-07
Iteration Output

### LINKED ISSUE #301: Output from iterations
Update the "Output" tab to accommodate for showing the output of each iteration, when iterating Gudrun.

jswift-stfc:
PR to show output of multiple iterations in the "Output" tab. Closes #301.

### DIFF
--- a/src/gui/widgets/main_window.py
+++ b/src/gui/widgets/main_window.py
@@ -1038,21 +1038,25 @@ class GudPyMainWindow(QMainWindow):
             self.nextIterableProc()
 
     def nextIteration(self):
-        self.outputIterations[self.currentIteration+1] = self.output
-        self.output = ""
         if self.error:
             self.proc.finished.connect(self.procFinished)
         if isinstance(self.iterator, TweakFactorIterator):
             self.iterator.performIteration(self.currentIteration)
             self.gudrunFile.write_out()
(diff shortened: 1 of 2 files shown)


trisyoungs (changes requested):
In principle this works completely as expected, but I have one question about the iterative procedure itself.  When performing the iterative w/l subtraction multiple times I seem to accumulate on top of past iterations.  This is one acceptable behaviour, but often (at least by default) the memory of previous iterations should be erased.

Regarding display of the items in the "Output" tab, I think there may be a case for using a tree rather than a list.  That way, all the output from an individual iteration could be visually grouped together. You could also offer the user an option to partition the tree by sample instead of iteration, showing all the different iterations as children.

Finally, a layout request!  The iterations dialog is a bit clunky and given the addition of more iterators in future will require some re-imagining. Perhaps a tabbed view with each iterator and its options displayed on a separate page?

These last two modifications could be addressed as separate PRs

trisyoungs:
Also, in the re-imagined iteration dialog the inelasticity subtraction option should definitely be the default visible one!

trisyoungs:
And just noticed another thing - the statusbar starts out by saying "Inelasticity subtractions 0/3", but most of our users aren't zero-indexed... :)

PR #313, opened 2022-02-17
Reorganise output tab

### LINKED ISSUE #309: Reorganise output tab for iterations
It's probably better to use a `QTreeView` rather than  a `QTreeListView` for the output navigation widget, especially in the case of iterations. This could also have the option to partition by sample/iteration.

jswift-stfc:
The output tab has been reorganised to use a `QTreeView` for navigation of output. When more than 1 iteration is performed, the tree gains parent nodes with the different outputs as leaf nodes, for each iteration. During single iterations / standard runs of `gudrun_dcs`, all nodes are leaf nodes, essentially the tree is flattened into a `QListView` style widget.

Closes #309.


trisyoungs (approved):
This works well and really helps the user navigate through the output (if they need to!).  I ran a two-iteration inelasticity subtraction as a test on my slow Windows machine and it felt as though nothing was happening for a while!  I wonder if its possible to implement incremental updating of the output as the calculation proceeds, rather than waiting until the end before populating the output tree?

jswift-stfc:
> 

I'll add an issue to address that in a separate PR!

PR #317, opened 2022-03-10
Iteration workflows

### LINKED ISSUE #117: Broader set of tools for iteratively tweaking values
Currently, we can only iteratively tweak the `tweak factor`. It would be useful if we could iteratively tweak other values, such as the `density`, `thickness/radius` (depending on geometry) and `composition`. Since these values are not always known.

### LINKED ISSUE #310: Refactor iterations dialog
With the number of iterators growing, the iteration dialog is becoming clunky. We could separate iterators by tabs. Also the Wavelength Subtraction Iterator should be the default mode of iteration.

jswift-stfc:
PR to implement additional iteration workflows.

- Iteration by thickness, for flatplate samples.
- Iteration by radii, for cylindrical samples.
- Iteration by density.
- Iteration by tweaking ratio of component(s).

Potentially closes #117?

Edit: closes #310, #318.


trisyoungs (approved):
Light testing complete, and all appear to work as expected. Structure of the iterators with the overloaded "modification" function is nice.  Some of the longer iterators perhaps lack a little bit of commenting?  Finally, the iterator selection dialog - the point size of the text explaining each iteration type seems to be a little bigger than elsewhere in the GUI.

PR #323, opened 2022-03-23
Run Gudrun after Purging

### LINKED ISSUE #320: Gudrun doesn't run after purge
Running Gudrun when a purge hasn't been detected / performed raises the dialog to run the purge as expected. However, after selecting "Purge with default parameters" one would expect Gudrun to run once the purge has completed, but it does not.

jswift-stfc:
Small PR which fixes bug described in #320.

gudrun_dcs is automatically run after a purge is performed, when the purge was executed via a prompt (i.e. the prompt which suggests that detectors have not yet been purged).

Closes #320.

### DIFF
--- a/src/gui/widgets/main_window.py
+++ b/src/gui/widgets/main_window.py
@@ -1018,8 +1018,8 @@ class GudPyMainWindow(QMainWindow):
                     self.mainWidget,
                     "GudPy Error", "Couldn't find purge_det binary."
                 )
-            self.setControlsEnabled(True)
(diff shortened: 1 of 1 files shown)


trisyoungs (approved):
Gudrun now correctly runs after the purge has completed, however, the label on the status indicator reads "No task running" when Gudrun is doing its thing.

jswift-stfc:
> 

What platform was this on? I can't seem to replicate on Linux.

PR #327, opened 2022-04-04
Component / Composition bug fixes

### LINKED ISSUE #324: Modifying Component doesn't trigger setModified
When modifying (adding/removing/altering) components, the setModified method isn't triggered, so the GUI seems as if the input file hasn't been modified, and hence, you cannot "Save".

jswift-stfc:
PR to fix various bugs with the Component / Composition models/tables. Closes #324.


trisyoungs (approved):
Seems sensible enough!

PR #341, opened 2022-05-09
Modulation Excitation

### LINKED ISSUE #51: Read/Write Nexus file
Implement a class, `NexusFile`, which allows parsing, reading and writing of files with the .nxs extension.

jswift-stfc:
PR implementing a Modulation Excitation processing pipeline into `GudPy` - this interfaces with `ModEx`.

There are 2 main ways that Modulation Excitation can be performed:

1) By defining a "Period" with "Pulses", and choosing an event to consider as the beginning of the first pulse. An extrapolation mode is also selected.
2) By choosing a spectra, and using all events as "Pulses", defining a "Period Duration" which is considered as the length of the pulse.

Further options are supplied, such as deciding whether to use a temporary directory whilst performing processing (thus rendering the outputted NeXus files inaccessible), or using a `writable` data file directory. In the latter case, the minimised, pulse-bounded NeXus files will be available afterwards. Also, the option to "interpolate" the produced mint files is offered.

Closes #51.

EDIT:

This functionality has been generalised to "Nexus Processing". This is only accessible by setting environment variables - `NEXUS_PROCESSING_ENABLED` and `SINGULARITYENV_NEXUS_PROCESSING_ENABLED` for general usage and usage through singularity, respectively.


trisyoungs (commented):
This is unavoidably a monster PR, so the most relevant reviewing will be a thorough test of the code (which may have to wait until next week).

One comment on comments.  The new modulation excitation routines lack your usual high level of commenting.  For future maintenance it would be nice to have (at the least) some function explanations added there.

trisyoungs (approved):
I like it! :+1:

PR #346, opened 2022-05-18
Display Packing Fraction

### LINKED ISSUE #8: Rename tweak factor appropiately
The 'tweak factor' is not descriptive, and can be renamed. It has been suggested that it is renamed to 'packing fraction'.

jswift-stfc:
This PR potentially solves one of the original "issues" - #8. Whilst renaming the 'tweak factor' to 'packing fraction', and performing calculations during reading/writing etc. could be potentially dangerous, with newer/older input file formats, it doesn't do any harm to simply display the packing fraction. Hence, this PR implements an extra exponential spin box next to the Tweak Factor (for both samples and containers), this is updated as the tweak factor changes, and vice versa - however, the packing fraction is not stored as a class attribute.

This is just an idea - and certainly does not need to be merged in, if it's not an appropriate enhancement.

Closes #8.

### DIFF
--- a/gudpy/gui/widgets/slots/container_slots.py
+++ b/gudpy/gui/widgets/slots/container_slots.py
@@ -90,15 +90,18 @@ class ContainerSlots():
         self.widget.containerCrossSectionFileWidget.setVisible(
             self.container.totalCrossSectionSource == CrossSectionSource.FILE
         )
-        # Set the tweak factor and packing fraction (reciprocal of tweak factor).
+        # Set the tweak factor and packing fraction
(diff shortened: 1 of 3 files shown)


trisyoungs (approved):
I haven't been able to test this "live", but I think this is a sensible addition, so as long as you're happy it does what it should I'm happy for it to be merged in.

jswift-stfc:
> I haven't been able to test this "live", but I think this is a sensible addition, so as long as you're happy it does what it should I'm happy for it to be merged in.

Okay! Yeah, it was more of a conceptual approval I was looking for :)

PR #347, opened 2022-05-18
Qt kernel fix

rprospero:
This should fix the kernel issue with Qt6 on CentOS 7


trisyoungs:
@jswift-stfc for context, IDAaaS have decided to regress to a Linux kernel from 2015 (best not to ask why...) which introduces a library error with Qt which depends on a kernel function not introduced until, maybe, 2016.  So, Adam has worked a solution for Dissolve, and has applied it to GudPy in this PR.  I'll let you review and approve, then I'll do the same and merge it in.

jswift-stfc (approved):
Working as expected - thank you for doing this @rprospero!

trisyoungs (approved):

PR #359, opened 2022-05-26
Strict Composition Table

jswift-stfc:
This PR makes the composition tables throughout the GUI stricter.

- When a mass number is entered, the atomic symbol is used to determine if this is a valid isotope. If it is not, then the mass number is not applied.
- When an atomic symbol is entered, the mass number is used to determine if this is a valid isotope. If it is not, then the mass number is set to zero (natural).

The main purpose for this PR is that when invalid isotopes are entered, it breaks the expected DCS level calculations.

Doesn't really need testing or code review, just looking for conceptual feedback!

### DIFF
--- a/gudpy/core/isotopes.py
+++ b/gudpy/core/isotopes.py
@@ -424,6 +424,11 @@ class Sears91():
             if self.element(isotope) == element
         ]
 
+    def findIsotope(self, element, mass):
+        for isotope in self.isotopes(element):
+            if self.mass(isotope) == mass:
+                return isotope
+
     def isIsotope(self, element, mass):
         isotopes = self.isotopes(element)
         for isotope in isotopes:
--- a/gudpy/gui/widgets/tables/composition_delegate.py
+++ b/gudpy/gui/widgets/tables/composition_delegate.py
@@ -1,0 +1,95 @@
+from PySide6.QtWidgets import QLineEdit, QComboBox
+from PySide6.QtCore import Qt
+from core.isotopes import Sears91
+from gui.widgets.core.exponential_spinbox import ExponentialSpinBox
+from gui.widgets.tables.gudpy_tables import GudPyDelegate
+
+
+class CompositionDelegate(GudPyDelegate):
+    """
+    Class to represent a CompositionDelegate. Inherits GudPyDelegate.
+
+    ...
+
+    Methods
+    -------
+    createEditor(parent, option, index)
+        Creates an editor.
+    setEditorData(editor, index)
+        Sets data at a specific index inside the editor.
+    setModelData(editor, model, index)
+        Sets data at a specific index inside the model.
+    """
+
+    def createEditor(self, parent, option, index):
+        """
+        Creates an editor, and returns it.
+        Parameters
+        ----------
+        parent : QWidget
+            Parent widget.
+        option : QStyleOptionViewItem
+            Option.
+        index : QModelIndex
+            Index in to create editor at.
+        Returns
+        -------
+        QLineEdit | QSpinBox | QDoubleSpinBox
+            The created editor.
+        """
+        col = index.column()
+        if col == 0:
+            editor = QLineEdit(parent)
+        elif col == 1:
+            editor = QComboBox(parent)
+            sears91 = Sears91()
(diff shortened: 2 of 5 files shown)


gudpy/gui/widgets/tables/composition_table.py
```diff
@@ -75,13 +76,31 @@ def setData(self, index, value, role):
         row = index.row()
         col = index.column()
         if role == Qt.EditRole:
+            sears91 = Sears91()
             if col == 0:
                 if value == "D":
                     self._data[row].atomicSymbol = "H"
                     self._data[row].massNo = 2
                     return True
```
trisyoungs, line 84: Returning here won't emit the `dataChanged` signal.

trisyoungs (approved):
Seems like a sensible improvement.  Dissolve takes this strictness one step further and uses a combobox delegate on the isotope data so that only valid isotopes can be selected.  May be worth considering if it isn't a massive pain to do in Python?

jswift-stfc:
> 

Also, I really like the periodic table style selection that Dissolve has. Would it be overkill to implement that here?

gudpy/gui/widgets/tables/composition_table.py
jswift-stfc: Good spot!

trisyoungs:
I think the periodic table might be overkill here. People are so used to just typing in their elements by hand that this could even end up being an annoyance!

gudpy/core/isotopes.py
```diff
@@ -424,6 +424,11 @@ def isotopes(self, element):
             if self.element(isotope) == element
         ]
 
+    def findIsotope(self, element, mass):
+        for isotope in self.isotopes(element):
+            if self.mass(isotope) == mass:
+                return isotope
```
trisyoungs, line 430: Super-dumb question here (it's very late...) but don't you need a default `return` if the isotope isn't found?

gudpy/gui/widgets/tables/composition_delegate.py
```diff
@@ -0,0 +1,104 @@
...
+        """
+        Creates an editor, and returns it.
+        Parameters
+        ----------
+        parent : QWidget
+            Parent widget.
+        option : QStyleOptionViewItem
+            Option.
+        index : QModelIndex
+            Index in to create editor at.
```
trisyoungs, line 34:
```suggestion
            Index to create editor at.
```

gudpy/gui/widgets/tables/composition_delegate.py
```diff
@@ -0,0 +1,104 @@
...
+            editor = ExponentialSpinBox(parent)
+        return editor
+
+    def setEditorData(self, editor, index):
+        """
+        Sets data at a specific index inside the editor.
+        Parameters
+        ----------
+        editor : QWidget
+            The editor widet.
```
trisyoungs, line 70:
```suggestion
            The editor widget.
```

gudpy/gui/widgets/tables/composition_delegate.py
```diff
@@ -0,0 +1,104 @@
...
+        elif index.column() == 2:
+            editor.setValue(value)
+
+    def setModelData(self, editor, model, index):
+        """
+        Sets data at a specific index inside the model.
+        Parameters
+        ----------
+        editor : QWidget
+            The editor widet.
```
trisyoungs, line 88:
```suggestion
            The editor widget.
```

gudpy/test/test_models.py
```diff
@@ -154,7 +160,8 @@ def testCompositionModel(self):
             model.data(model.index(0, 0, QModelIndex()), Qt.EditRole), "V"
         )
         self.assertEqual(
-            model.data(model.index(0, 1, QModelIndex()), Qt.EditRole), 0
+            model.data(model.index(0, 1, QModelIndex()), Qt.EditRole),
+            self.sears91.isotope(self.sears91.findIsotope("V", 0))
```
trisyoungs, line 164: Should we add a companion "negative" test here, searching for an isotope that doesn't actually exist?

trisyoungs (approved):

PR #361, opened 2022-06-07
Batch Processing

### LINKED ISSUE #116: Batch Processing
Rename the current `RunIndividualFiles` class to something along the lines of `BatchProcessFiles`, and add the option to specify a `batch_size`, to enable batch processing. Then instead of running data files in batches of size 1, as in `RunIndividualFiles`,  run data files in batches of size `batch_size`. `RunIndividualFiles` can simply specify a `batch_size` of 1.

jswift-stfc:
Another large feature PR!

This implements Batch Processing, there are a few ways that this can work.
 - Raw batch processing using a given batch size.
 - Raw batch processing using a given batch size and step size (this defaults to the batch size).
 - Iterative batch processing using a given batch size, (step size), iteration mode and maximum number of iterations, iteratively processes each batch.
 - Iterative batch processing using a given batch size, (step size), iteration mode, convergence tolerance and a  maximum number of iterations, iteratively processes each batch, using the convergence tolerance for an early return.

The organisation of output files is consistent with the other workflows. Additionally, a diagnostic file is produced which gives insight into the contents of each batch, as well as resultant values (when iterating), and the percentage error in the DCS level.

Closes #116.

### DIFF
--- a/gudpy/gui/widgets/slots/output_slots.py
+++ b/gudpy/gui/widgets/slots/output_slots.py
@@ -4,7 +4,7 @@ class OutputSlots():
         self.widget = widget
         self.parent = parent
 
-    def setOutput(self, output, task):
+    def setOutput(self, output, gudrunFile, task):
         self.output = output
         self.task = task
-        self.widget.outputTree.buildTree(self.parent.gudrunFile, output, self)
+        self.widget.outputTree.buildTree(output, gudrunFile, self)
--- a/README.md
+++ b/README.md
@@ -67,3 +67,14 @@ More advanced iteration workflows
  - In the case of a single component it balances the ratio of that component against the others in the composition.
  - For two components, it balances the ratios of the two against eachother, without altering the total number of molecules in the composition.
  - Since this can be a particularly slow workflow, a convergence tolerance can be defined.
+### Batch Processing
+GudPy includes a Batch Processing pipeline, that is flexible and can interface with basic iteration workflows.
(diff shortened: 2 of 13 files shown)


gudpy/gui/widgets/slots/output_slots.py
```diff
@@ -4,7 +4,7 @@ def __init__(self, widget, parent):
         self.widget = widget
         self.parent = parent
 
-    def setOutput(self, output, task):
+    def setOutput(self, output, gudrunFile, task):
         self.output = output
         self.task = task
-        self.widget.outputTree.buildTree(self.parent.gudrunFile, output, self)
+        self.widget.outputTree.buildTree(output, gudrunFile, self)
```
trisyoungs, line 10: I notice that the order of the parameters has changed here - just checking that this isn't an error?

trisyoungs (approved):
:+1: Approving for merge in principle as code looks OK, but I have not tested it at all!  Perhaps one to promote next week and get feedback on.

jswift-stfc:
> 👍 Approving for merge in principle as code looks OK, but I have not tested it at all! Perhaps one to promote next week and get feedback on.

No worries! I'll merge in for now. Tom said he was going to test against your old multi-gud stuff, so merging in now and getting this on IDAaaS should make that easier for him. I'll open an issue to add unit tests for this in future (when we can actually confirm it is working as expected!).

PR #377, opened 2022-07-04
Continue iterations batch processing

### LINKED ISSUE #372: Continue Iterations (Batch Processing)
Currently, when using the batch processing pipeline, and iterating, each batch is **iterated** in isolation (hence, all batches are run in the sample call to `gudrun_dcs`. We need to add an option to "continue iterations"; this should enable the final value of iterating the current batch to be applied to the next batch - hence, the need to ***run*** batches in isolation, to achieve communicative iterations.

jswift-stfc:
Implements "propogation of first batch" during the batch processing pipeline.

Closes #372.

### DIFF
--- a/gudpy/core/run_batch_files.py
+++ b/gudpy/core/run_batch_files.py
@@ -1,41 +1,64 @@
+from cmath import e
 from copy import deepcopy
+from mimetypes import init
 import os
+from numpy import maximum
+
+from requests import head
 
 from core.enums import IterationModes
 from core.tweak_factor_iterator import TweakFactorIterator
 from core.thickness_iterator import ThicknessIterator
 from core.radius_iterator import RadiusIterator
 from core.density_iterator import DensityIterator
+from gudpy.core.enums import ITERATION_MODES
 
 
 class BatchProcessor:
     def __init__(self, gudrunFile):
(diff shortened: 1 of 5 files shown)


gudpy/core/run_batch_files.py
```diff
@@ -68,17 +85,56 @@ def writeDiagnosticsFile(self, path, batch, iterationMode):
...
+                    sampleB.sampleTweakFactor = sampleA.sampleTweakFactor
+                elif iterationMode == IterationModes.THICKNESS:
+                    sampleB.upstreamThickness = sampleA.upstreamThickness
+                    sampleB.downstreamThickness = sampleA.downstreamThickness
+                elif iterationMode == IterationModes.INNER_RADIUS:
+                    sampleB.innerRadius = sampleA.innerRadius
+                elif iterationMode == IterationModes.OUTER_RADIUS:
+                    sampleB.outerRadius = sampleA.outerRadius
+                elif iterationMode == IterationModes.DENSITY:
+                    sampleB.desity = sampleA.density
```
trisyoungs, line 101:
```suggestion
                    sampleB.density = sampleA.density
```

trisyoungs (approved):
Haven't tested, but looks good in principle!

PR #395, opened 2022-10-01
Fix inelasticity iterations

### LINKED ISSUE #386: Inelasticity correction not working?
Iterating GudPy with the inelasiticy correction appears to generate the correct output structure (e.g. directories named after the iterations) but the correction itself does not seem to be subtracted from the data in the plots. May be due to an issue with "Subtract Wavelength Binned Data" not being auto-enabled when running this iterator?

jws-1:
I think that this PR fixes the issue outlined in #386, however I am not entirely sure - it needs to be tested and verified. Will close #386.


trisyoungs:
Obsoleted by #406.

PR #396, opened 2022-10-01
Fix broken X-Scale

### LINKED ISSUE #382: Step size incorrect when reading from yaml
The X-scale "Step Size" value does not appear to be set correctly when reading in a yaml input file.  Previously saving the file with log binning checked and a step size of 0.03 entered, re-loading the file reverts to a value of -0.01 and log binning _unchecked_.

jws-1:
I believe the issue outlined in #382 was related to two different issues:
 - Changing the XMin never worked, due to a typo.
 -  Switching between scales was being handled incorrectly leading to incorrectly saving values.
 - This wasn't noticed, in this case, when saving to a text file because that stores the XStep as -0.01 when logarithmic binning is checked.
 

So, I believe this closes #382.

### DIFF
--- a/gudpy/gui/widgets/slots/instrument_slots.py
+++ b/gudpy/gui/widgets/slots/instrument_slots.py
@@ -962,8 +962,8 @@ class InstrumentSlots():
         [
             widget.setEnabled(state) for widget in self.scales[Scales.Q][1:]
         ]
-        if state:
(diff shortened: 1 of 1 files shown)


trisyoungs:
Superseded by #401 (duplication of work to regenerate CI artifacts).

PR #402, opened 2022-11-18
fix: use platform specific path separators. JS reimplimented

TerriLouise:
PR should fix issues with windows build


trisyoungs (approved):
Lightly tested on Windows - appears to work!

:+1:

PR #411, opened 2023-01-13
Fix publishing logic

trisyoungs:
A quick PR to implement some changes to the logic used in the `if` expressions in the publishing action, after finding with JV2 that it didn't work as previously written!


trisyoungs:
@TerriLouise - Finally, a big green tick! :)

TerriLouise (approved):

PR #427, opened 2023-07-10
Sorts data files in numerical order

### LINKED ISSUE #420: Data files automatically sorted
Sample data files aren't currently sorted numerically if not pre-sorted during selection.
image

rhinoella:
Closes #420

### DIFF
--- a/gudpy/core/gudrun_file.py
+++ b/gudpy/core/gudrun_file.py
@@ -613,6 +613,8 @@ class GudrunFile:
             dataFiles = []
             for _ in range(numberOfFiles):
                 dataFiles.append(firstword(self.getNextToken()))
+            # Reverses list so that it is in ascending order
+            dataFiles.reverse()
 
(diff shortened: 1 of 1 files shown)


gudpy/core/gudrun_file.py
```diff
@@ -613,6 +613,8 @@ def parseNormalisation(self):
             dataFiles = []
             for _ in range(numberOfFiles):
                 dataFiles.append(firstword(self.getNextToken()))
+            # Reverses list so that it is in ascending order
+            dataFiles.reverse()
```
trisyoungs: Since there is no guarantee of the ordering of the files in the input list, is `reverse()` reliable here? Don't we want to `sort()` instead?

rhinoella: Yes, this is very true! I changed it to `sort()`

trisyoungs (approved):

PR #433, opened 2023-07-12
Improved error handling

### LINKED ISSUE #432: Application becomes unresponsive when incorrect file is sent into NeXus Processing
If we set data file type as .nxs and point to a data file directory that does not contain a nexus file the application will crash.

rhinoella:
- File existence checking done before Nexus Processing dialog opens (stops application from crashing)
- Option to change data file types is disabled as this can cause users to set an inconsistent file type and cause a crash
- Users can set data file type when creating a new input file (it used to restrict it to raw only)

Closes #432

### DIFF
--- a/gudpy/gui/widgets/core/main_window.py
+++ b/gudpy/gui/widgets/core/main_window.py
@@ -693,7 +693,9 @@ class GudPyMainWindow(QMainWindow):
             self.gudrunFile.instrument = GudrunFile(
                 configurationDialog.configuration, config_=True
             ).instrument
-            self.gudrunFile.instrument.dataFileType = configurationDialog.dataFileType
+            self.gudrunFile.instrument.dataFileType = (
+                configurationDialog.dataFileType
+            )
             self.updateWidgets()
 
     def updateFromFile(self):
--- a/gudpy/gui/widgets/core/main_window.py
+++ b/gudpy/gui/widgets/core/main_window.py
@@ -693,6 +693,7 @@ class GudPyMainWindow(QMainWindow):
             self.gudrunFile.instrument = GudrunFile(
                 configurationDialog.configuration, config_=True
             ).instrument
+            self.gudrunFile.instrument.dataFileType = configurationDialog.dataFileType
             self.updateWidgets()
 
     def updateFromFile(self):
--- a/gudpy/gui/widgets/core/main_window.py
+++ b/gudpy/gui/widgets/core/main_window.py
@@ -641,7 +641,6 @@ class GudPyMainWindow(QMainWindow):
                 self.gudrunFile = GudrunFile(path=path)
                 self.updateWidgets()
                 self.mainWidget.setWindowTitle(self.gudrunFile.path + " [*]")
-
             except ParserException as e:
                 QMessageBox.critical(self.mainWidget, "GudPy Error", str(e))
 
--- a/gudpy/gui/widgets/core/main_window.py
+++ b/gudpy/gui/widgets/core/main_window.py
@@ -641,6 +641,7 @@ class GudPyMainWindow(QMainWindow):
                 self.gudrunFile = GudrunFile(path=path)
                 self.updateWidgets()
                 self.mainWidget.setWindowTitle(self.gudrunFile.path + " [*]")
+
             except ParserException as e:
                 QMessageBox.critical(self.mainWidget, "GudPy Error", str(e))
 
@@ -1257,6 +1258,9 @@ class GudPyMainWindow(QMainWindow):
         )
 
     def nexusProcessing(self):
+        if not self.checkFilesExist_():
+            return
+
         self.setControlsEnabled(False)
         nexusProcessingDialog = NexusProcessingDialog(
             self.gudrunFile, self.mainWidget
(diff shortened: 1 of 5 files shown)


gudpy/gui/widgets/dialogs/configuration_dialog.py
```diff
@@ -96,6 +108,14 @@ def setConfiguration(self, _prev, _curr):
             open(self.configuration, "r").readlines()[-1]
         )
 
+    def handleDataFileTypeChanged(self, index):
+        """
+        Sets file type upon change
```
jws-1:
```suggestion
        Update data file type
```

gudpy/gui/widgets/dialogs/configuration_dialog.py
```diff
@@ -55,6 +56,17 @@ def initComponents(self):
...
+        # Sets the combo box options
+        dataFileTypes = ["raw", "sav", "txt", "nxs", "*"]
+        self.widget.dataFileTypeCombo.addItems(dataFileTypes)
+        self.widget.dataFileTypeCombo.setCurrentIndex(
+            dataFileTypes.index(self.dataFileType)
+        )
+
+        self.widget.dataFileTypeCombo.currentIndexChanged.connect(
+            self.handleDataFileTypeChanged
+        )
```
jws-1: Might be better to use `currentTextChanged`

gudpy/gui/widgets/dialogs/configuration_dialog.py
```diff
@@ -96,6 +108,14 @@ def setConfiguration(self, _prev, _curr):
             open(self.configuration, "r").readlines()[-1]
         )
 
+    def handleDataFileTypeChanged(self, index):
+        """
+        Sets file type upon change
+        """
+        self.dataFileType = self.widget.dataFileTypeCombo.itemText(
+            index
+        )
```
jws-1: If we use `currentTextChanged`, then we could directly update the `dateFileType`

gudpy/gui/widgets/core/main_window.py
```diff
@@ -641,6 +641,7 @@ def loadInputFile_(self):
                 self.gudrunFile = GudrunFile(path=path)
                 self.updateWidgets()
                 self.mainWidget.setWindowTitle(self.gudrunFile.path + " [*]")
+
```
jws-1:
```suggestion
```

gudpy/gui/widgets/slots/instrument_slots.py
```diff
@@ -31,6 +31,17 @@ def setInstrument(self, instrument):
             dataFileTypes.index(self.instrument.dataFileType)
         )
 
+        # Enable Nexus Processing if data file type is nxs
+        self.widget.nexusDefintionFileLineEdit.setEnabled(
+            self.instrument.dataFileType in ["nxs", "NXS"]
+        )
+        self.widget.browseNexusDefinitionButton.setEnabled(
+            self.instrument.dataFileType in ["nxs", "NXS"]
+        )
```
jws-1: We also do this a bit later on (I think lines 167-172)

jws-1 (changes requested):
Great, I like it! Just a couple of very small comments.

It might be nice to add an `Enum` for the data files types (see core/enums.py), although this will require quite a lot of changes throughout the codebase so it's probably better off saved for another PR.

gudpy/gui/widgets/slots/instrument_slots.py
rhinoella: Good catch! Fixed it :)

gudpy/gui/widgets/dialogs/configuration_dialog.py
trisyoungs: :+1: for this suggestion. But capture its conversion to an Enum as a separate issue.

trisyoungs (commented):
I can think of a situation where these modifications will make things difficult, and that's where we have setup a processing with one file type (e.g. raw) and want to try/test the same processing with something else (probably nxs). We could capture this as a separate issue, and add a menu item to allow switching to a different file type _en masse_, switching file extensions for all loaded files automatically.

gudpy/gui/widgets/dialogs/configuration_dialog.py
rhinoella: Ahh I see, I changed it! :smile:

jws-1:
> I can think of a situation where these modifications will make things difficult, and that's where we have setup a processing with one file type (e.g. raw) and want to try/test the same processing with something else (probably nxs).

Hah. That'd be my fault for saying yesterday that I doubt someone would want to switch between data file types!

jws-1 (approved):
:+1:

trisyoungs:
> > I can think of a situation where these modifications will make things difficult, and that's where we have setup a processing with one file type (e.g. raw) and want to try/test the same processing with something else (probably nxs).
> 
> Hah. That'd be my fault for saying yesterday that I doubt someone would want to switch between data file types!

Well, *I* would want to, and it's all about me! :smiley:

trisyoungs (approved):

rhinoella:
> I can think of a situation where these modifications will make things difficult, and that's where we have setup a processing with one file type (e.g. raw) and want to try/test the same processing with something else (probably nxs). We could capture this as a separate issue, and add a menu item to allow switching to a different file type _en masse_, switching file extensions for all loaded files automatically.

I will re-enable the file type combo box, I just realized that the error checking bit I added stops the user from running processes with an inconsistent file type anyways so I don't have to disable this, I will just have to improve the error dialog. Does this address your concern?

PR #438, opened 2023-08-09
Updated Nexus Processing Dialog

### LINKED ISSUE #418: Default button on NeXuS processing dialog
The `OK` button is set as the default button in the NeXuS processing dialog, meaning that whenever another control (spinbox, table cell) is activated by pressing `Enter` after typing a value the `OK` button is activated also, leading to unexpected behaviour.

rhinoella:
The previous dialog was too large and would not scale well on smaller screens.
This PR also removes the default buttons that causes #418

Old layout:
image

New layout:
image

Closes #418

### DIFF
--- a/gudpy/core/nexus_processing.py
+++ b/gudpy/core/nexus_processing.py
@@ -28,6 +28,7 @@ class RawPulse:
     end : float
         End time of the pulse.
     """
+
     def __init__(self, start, end):
         """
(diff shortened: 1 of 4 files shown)


trisyoungs (approved):
This looks better to me.  There are probably still some space savings you can make, but those can be saved for a later date.

PR #439, opened 2023-08-09
Fixing unsafe comparisons that caused lint checks to fail

rhinoella:
Caused PR #438 checks to fail:

https://github.com/disorderedmaterials/GudPy/actions/runs/5810507244/job/15751562972#step:4:125


jws-1 (approved):
:+1:

PR #442, opened 2023-08-24
Removing type comparisons

rhinoella:
Replaced type comparisons to `isinstance()` in `chartview.py` that causes lint checks to fail.


rhinoella:
Closed this PR because unrelated commits were included in this branch for some reason

PR #443, opened 2023-08-24
Changed type comparisons in chartview.py

rhinoella:
Had to do this again because the lint checks were still unhappy. Changed type comparisons to `isinstance()` calls.


gudpy/gui/widgets/charts/chartview.py
```diff
@@ -152,7 +154,7 @@ def mouseReleaseEvent(self, event):
                     self.rubberBandOrigin.x(),
                     self.rubberBandOrigin.y(),
                     width,
-                    height
+                    height,
```
trisyoungs, line 157: Every fibre of my being wants to say that there shouldn't be a trailing comma on the last item, but I'm sure there's a good reason for it!

trisyoungs (approved):
Looks good!

PR #444, opened 2023-08-30
440 Define starting NeXus pulse

### LINKED ISSUE #440: Manually define starting pulse for NeXus processing
Being able to manually define a starting pulse for NeXus processing would be very useful - particularly in cases where there seem to be no pulses in the spectra, but the starting time of some pulse is known.

jws-1:
This PR implements the ability  to manually define a starting pulse for the NeXus processing to begin from, and therefore closes #440.

I'd suggest that this PR remains a draft until #438 goes in, and then the user-facing changes of this PR should then be reconsidered.

### DIFF
--- a/gudpy/gui/widgets/dialogs/nexus_processing_dialog.py
+++ b/gudpy/gui/widgets/dialogs/nexus_processing_dialog.py
@@ -108,7 +108,6 @@ class NexusProcessingDialog(QDialog):
         self.widget.lowerSpecSpinBox.setValue(min(spectra))
         self.widget.upperSpecSpinBox.setValue(max(spectra))
 
-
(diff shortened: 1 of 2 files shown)


trisyoungs (commented):
Intent looks good!

trisyoungs (commented):
Functionality seems fine, but if I'm honest I don't like the way the UI is implemented at present. I'd rather the controls were always visible, rather than being shown/hidden.  I would also lose the group boxes and tighten the margins:
image
I can push this mod if you want to save a bit of time, although the widget logic isn't quite right!

rhinoella:
This PR is now part of #457

rhinoella:
Closing due to Nexus Processing being removed from GudPy

PR #446, opened 2023-09-15
Improved safety of NeXus Processing temp directory

rhinoella:
Small change to how the temporary directory is managed in NeXus processing to improve safety.

- New temporary directory is created with each NeXus processing run
- Omits need for `shutil.rmtree'
- Deletes directory once NeXus run is complete

### DIFF
--- a/gudpy/core/nexus_processing.py
+++ b/gudpy/core/nexus_processing.py
@@ -252,7 +252,7 @@ class NexusProcessing:
         self.sample = None
         self.useTempDataFileDir = False
         self.interpolate = False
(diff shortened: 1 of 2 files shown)


trisyoungs (commented):
Nice little cleanup, but we need to take a bit of care here. The use of temporary vs "writable data directory" and the additional "output" directory is pretty confusing and could do with a larger scale tidy up.  If we delete the contents of tmp immediately after the processing has finished we will leave the user with nothing to work with...

rhinoella:
Closed due to Nexus Processing being removed from GudPy - #478

PR #448, opened 2023-09-18
Fix Gudrun / Purge Behaviour

trisyoungs:
One-liner PR to fix the situation where a purge had already been run, and the use want's to continue to run Gudrun with the purge files already present.


rhinoella (approved):
👍

PR #456, opened 2023-09-25
Small change to how window is named

rhinoella:
The previous title would show the whole path:
image

Now shows 'GudPy' and just the input filename (similar to how it is done in Dissolve):
image


trisyoungs (approved):
:+1:

PR #471, opened 2023-11-24
Fix IDAaaS Setup Script

trisyoungs:
Just add a missing library and fix HDF5 compiler wrappers (for Gudrun) which suddenly broke.


rhinoella (approved):
👍

PR #473, opened 2023-11-24
Fixes November 2023

### LINKED ISSUE #466: File type selector has blank entry
When opening an input file the file type selector in the file dialog has three legit entries ("YAML", "Gudrun Compatible", and "Sample Parameters") and a fourth completely blank entry which should be removed.

### LINKED ISSUE #467: Detector calibration file not covered by file existence check
The "Check Files Exist" feature does not seem to cover the "Detector Calibration File", since putting in a bogus filename doesn't raise a warning in the dialog.

### LINKED ISSUE #468: Check files exist option should report positive results
If all files are found when running "Check Files Exist" there is no user feedback whatsoever - we really should just raise a simple dialog to inform the user that everything is OK.

### LINKED ISSUE #469: No indication that the GUI has been modified
The issue probably being that:
`QWidget::setWindowModified: The window title does not contain a '[*]' placeholder`

### LINKED ISSUE #472: Missing binary has odd effects on the GUI
If, e.g. the `gudrun_dcs` binary is missing and we try to Run -> Run Gudrun we get a nice error dialog saying that it can't be found. However, the File menu (except for the "Save" option) and the Run menu are then completely disabled afterwards.

trisyoungs:
A PR handling a few small fixes to improve the UX.

Closes #466.
Closes #467.
Closes #468.
Closes #469.
Closes #472.

### DIFF
--- a/gudpy/core/file_library.py
+++ b/gudpy/core/file_library.py
@@ -48,7 +48,9 @@ class GudPyFileLibrary:
 
         # Collect files of static objects
         self.files = {
-            "Detector Calibration File": gudrunFile.instrument.detectorCalibrationFileName,
+            "Detector Calibration File": (
+                gudrunFile.instrument.detectorCalibrationFileName
+            ),
             "Groups File": gudrunFile.instrument.groupFileName,
             "Deadtime Constants File": (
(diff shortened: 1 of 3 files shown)


rhinoella (approved):
Looks good!

PR #474, opened 2023-11-30
Display live output from running processes

rhinoella:
Currently GudPy only displays outputs once the running process is complete. This PR enabled live outputs of Gudrun and Purge which will aid in debugging.


trisyoungs (approved):
Tested - works nicely!

PR #478, opened 2024-01-17
Remove Nexus Processing

rhinoella:
NeXus processing is removed from GudPy as it is no longer part of the application.


trisyoungs (approved):

rhinoella:
Implemented in #487

Issue #457, opened 2023-09-26
Outputs Update


rhinoella:
Reform the way Gudrun is run and the way the outputs are handled.

- [x] Have all the output files stored and organised in a directory
- [x] Have Gudrun dcs always run in a temporary directory and copy the files out of it
- [x] Have purge run in a temporary directory and place purge files in project directory
- [x] Update Nexus Processing to not need the temporary/writable directory options
- [ ] Allow the project directory to be reloaded into Gudpy to continue work

PR #479, opened 2024-01-17
Load GudPy work from project folder

rhinoella:
Allows GudPy to be loaded from a previously outputted project folder. Supports the move from input files to project folders in #457.

- GudPy still supports loading from input files
- When an input file is loaded, when the user sets a save location, a project folder is created there and a new copy of the input file will exist there (GudPy will not overwrite existing input files outside of a project folder)


gudpy/core/gud_file.py
```diff
@@ -21,7 +21,7 @@ class GudFile:
     ----------
     path : str
         Path to the file.
-    outpath : str
+    OUTPATH : str
```
trisyoungs, line 24: Just an innocent query - why the capitalisation for this member var?

gudpy/gui/widgets/core/main_window.py
```diff
@@ -461,6 +464,8 @@ def initComponents(self):
 
         self.mainWidget.loadInputFile.triggered.connect(self.loadInputFile_)
 
+        self.mainWidget.loadProject.triggered.connect(self.loadProject_)
+
```
trisyoungs, line 468: Another probably obvious question - why do these two functions have the `_` appended to their names?

gudpy/gui/widgets/core/main_window.py
```diff
@@ -600,26 +605,72 @@ def loadInputFile_(self):
...
 
+    def loadProject_(self):
+        """Load from previous GudPy project
+        """
+        projectDir = QFileDialog.getExistingDirectory(self, 'Select Project')
+
+        try:
+            self.gudrunFile = GudrunFile(
+                projectDir=projectDir, format=Format.YAML)
+            autosave = self.tryLoadAutosaved(self.gudrunFile.path)
```
trisyoungs, line 616: Perhaps the variable name could be a little more descriptive? Unless I'm misunderstanding the role here!
```suggestion
            autosavedFile = self.tryLoadAutosaved(self.gudrunFile.path)
```

gudpy/gui/widgets/core/main_window.py
```diff
@@ -629,7 +680,14 @@ def saveInputFileAs(self):
...
                     "GudPy Warning",
-                    f"Cannot save to {filename}, gudpy.txt is reserved.",
+                    f"Cannot export to {filename}, gudpy.txt is reserved.",
+                )
+                return
+            if os.path.dirname(filename) == self.gudrunFile.projectDir:
+                QMessageBox.warning(
+                    self.mainWidget,
+                    "GudPy Warning",
+                    "Do not modify project folder."
```
trisyoungs, line 690: I suppose this is to check whether the YAML gudrunFile exists in the project directory is was born in. If the user legitimately moves the project folder whilst reorganising files, would we be better off taking a softer approach ans popping up a message box a la "It looks like the project folder has been moved to a new location.  Would you like to update the main GudPy input file to match the new location?"

trisyoungs (approved):
Looking through this really made me appreciate what a nice idea this was in the first place. I fully believe this is how it should be.

Few minor comments (questions, really, for my own benefit!) to address before merging, but otherwise looks good.

gudpy/core/gud_file.py
rhinoella: Capitalisation means its intended to be a constant!

gudpy/gui/widgets/core/main_window.py
rhinoella: I believe it is to distinguish between the signal and the function

gudpy/gui/widgets/core/main_window.py
rhinoella: This is to not allow user's to export an input file into the project folder itself!

PR #480, opened 2024-01-24
477 Make GudPy compatible with GudrunGUI

### LINKED ISSUE #477: Gudrun files are not quite written in the correct format
In order to load a .txt formatted file written by GudPy into GudrunGUI one needs to make two corrections to the formatting:

1. The number of spaces between sample composition details and the text comment needs to be increased from 8 to 10. For example:
```
V  0  0.9485        Composition
```
...should be...
```
V  0  0.9485          Composition
```

2. An unused third parameter for the exponential values was removed some time ago, but is necessary for GudrunGUI not to choke on the file. For example:
```
0  1          Exponential amplitude and decay [1/Å]
```
...should be...
```
0  1  0          Exponential amplitude and decay [1/Å]
```

rhinoella:
Closes #477 

- Increases the space between the values and 'Composition' keyword to 10
- Appends a blank value to 'Exponential Values' (depreciated but still needed for GudrunGUI)

### DIFF
--- a/gudpy/core/composition.py
+++ b/gudpy/core/composition.py
@@ -231,7 +231,7 @@ class Composition():
         string = ""
         for el in self.elements:
             string += (
-                str(el) + config.spc5 +
+                str(el) + config.spc10 +
                 "Composition\n"
             )
 
--- a/gudpy/core/composition.py
+++ b/gudpy/core/composition.py
@@ -5,6 +5,7 @@ from core.element import Element
 from core.isotopes import Sears91
 from core.exception import ChemicalFormulaParserException
 from core.mass_data import massData
+from core import config
 
 
 class ChemicalFormulaParser():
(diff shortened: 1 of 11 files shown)


gudpy/core/composition.py
```diff
@@ -230,7 +231,7 @@ def __str__(self):
         string = ""
         for el in self.elements:
             string += (
-                str(el) + "        "
+                str(el) + config.spc5 +
```
trisyoungs: The variable `config.spc5` is rather confusingly named, don't you think, given that it has ten space in it! For sanity we might consider renaming it, but is it used elsewhere?

trisyoungs (approved):

gudpy/core/composition.py
rhinoella: Yes, it is, but I can change all references to it?

rhinoella: I think it is named spc5 because Gudrun reads two spaces as one?

trisyoungs: We shouldn't get bogged down in this, but in `config` there's also `spc2` which looked like just two spaces?

rhinoella: Renamed it! I agree that spc10 is better!

PR #482, opened 2024-01-25
470 Remove 'Duplicate File' Button & Select Multiple Files

### LINKED ISSUE #470: No need for duplicate file button for file lists
In all areas where we have the ability to specify target data (raw, nxs) files we have the add, remove, and duplicate options.  The latter option is not required and should be removed, as specifying the same file multiple times will likely lead to undefined behaviour within Gudrun itself, or in the resulting output.

Also, the ability to multi-select files and remove them would be nice!

rhinoella:
This PR removes the button that duplicates data files, as well as allows for multiple files to be selected and removed.

Closes #470

### DIFF
--- a/gudpy/gui/widgets/tables/data_file_list.py
+++ b/gudpy/gui/widgets/tables/data_file_list.py
@@ -40,4 +40,5 @@ class DataFilesList(QListView):
         Removes rows from the model.
         """
         if self.currentIndex().isValid():
-            self.model().removeRow(self.currentIndex().row())
+            for index in self.selectedIndexes():
+                self.model().removeRow(index.row())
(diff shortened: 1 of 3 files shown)


gudpy/gui/widgets/tables/data_file_list.py
```diff
@@ -40,4 +40,5 @@ def removeItem(self):
         Removes rows from the model.
         """
         if self.currentIndex().isValid():
-            self.model().removeRow(self.currentIndex().row())
+            for index in self.selectedIndexes():
+                self.model().removeRow(index.row())
```
trisyoungs: I'm not in a position to test this at present, but is this logic correct? Won't the `row()`s of every selected index after the first deletion be wrong?

rhinoella: Ah, yes I tested it further and it fails at times, so I have updated this line. Should work better now

trisyoungs (approved):
Bit clunky, but Qt doesn't exactly make this kind of the easy to do in a nice way!

PR #487, opened 2024-01-29
Create a GudPy class

rhinoella:
Very large PR that serves to reduce bugs and create a distinction between the headless and GUI modes of gudpy.

Key goals of this PR:
- Separate headless and GUI code
- Remove nexus processing
- Create a fully functioning headless mode
- Refactor to avoid needing multiple functions for the same purpose of running processes
- Improve safety in file management, run everything in context managers so that if there is an issue in execution the filesystem will not be modified
- Remove responsibilities from `GudPyMainWindow`
- Improve error handling and error reporting
- Implement the new project-system

Implements three significant modules:

## core/gudpy

The main backend module that handles all of the processing logic, made to be completely separate from Qt. Contains the following classes:

### GudPy
Class to act as the main GudPy backend. Holds all the necessary data and implements all functions needed to run GudPy headlessly.  

### Process
Base class for all processes. 

Implements a key function `_outputChanged` which is called on output of the underlying process, and allows the class to be subclassed by a QThread with this function to be overriden by a signal-emitting function- therefore removing the need to reimpliment the logic to be compatible with the GUI version.

### Purge(Process), Gudrun(Process)
Classes to manage the running of purge_det and gudrun_dcs binaries.

Runs all processes inside a temporary directory using a context manager to avoid polluting the user's directory. Replaces the use of `subprocess.run()` with `subprocess.Popen` to allow simpler live-output capture.

### GudrunIterator, CompositionIterator
Class to manage the iteration of gudrun_dcs.

Takes away logic/task queing responsibilities from the iteration dialog classes. Replaces using a queue with a list of Gudrun objects that are iterated over.

### RunModes
Class that combines the old `gudpy/core/run_containers_as_samples.py` and `gudpy/core/run_individual_files.py` modules into one class that can manage different modes to run gudrun.

### BatchProcessing
Class to manage the batch processing workflow, implementing all the newly refactored methods.

## gui/widgets/core/worker
The module that provides GUI compatibility/Multithreading to the classes in `core/gudpy`.

Subclasses `QThread` and the processing classes in `core/gudpy` to implement slots and signals and extra logic needed for the GUI. Allows for the headless mode to maintain complete separation from Qt.

Replaces the previous use of `QProcess` with creating `QThread` workers for each process.

## gui/widgets/core/control

Contains the class `GudPyController` to act as the controller between the GUI and the backend (GudPy) class. 
Significantly removes responsibility from `GudPyMainWindow` and creates a separation between logic and GUI methods.

Reimplements functions from `GudPy` class to accomodate the GUI.


gudpy/gui/widgets/dialogs/io.py
```diff
@@ -1,10 +1,75 @@
...
 from PySide6.QtCore import QFile
-from PySide6.QtWidgets import QDialog, QFileDialog, QMessageBox
 from PySide6.QtUiTools import QUiLoader
 
-from core.file_library import GudPyFileLibrary
+
+class MissingFilesDialog(QDialog):
+    """
+    Class to represent the MissingFilesDialog. Inherits QDialog.
+    This is the dialog window opened when a user wishes to iterate Gudrun.
```
trisyoungs: I guess this comment is a copy/paste "error" since missing files applies to all processes.

gudpy/core/gudpy.py
```diff
@@ -0,0 +1,980 @@
...
+
+    def gudrun(
+        self,
+        gudrunFile: GudrunFile,
+        purge: Purge = None,
+        iterator: iterators.Iterator = None
+    ) -> int:
+        self.checkBinary()
+        if not purge:
+            print("WARNING: Gudrun running without purge.")
```
trisyoungs: How effective is this warning message? Should we do something "stronger"?

gudpy/core/gudrun_file.py
```diff
@@ -915,6 +864,7 @@ def parseSample(self):
                 str(self.getNextToken()[:-2]).strip()
                 .replace("SAMPLE", "").strip()
             )
+            sample.name = sample.name.replace(" ", "_")
```
trisyoungs: Spaces are just one problem for sample naming - perhaps we should consider adding a general stripping function we can call, and which can be extended with other special characters as necessary.

gudpy/core/iterators.py
```diff
@@ -0,0 +1,806 @@
...
+            gudrunFile=gudrunFile,
+        )
+        return outputHandler.organiseOutput(exclude=exclude)
+
+
+class Radius(Iterator):
+    """
+    Class to represent a Radius Iterator. Inherits Iterator.
+    This class is used for iteratively tweaking the thickness of a
+    flatplate sample.
```
trisyoungs:
```suggestion
    This class is used for iteratively tweaking the radii of a
    cylindrical sample.
```

gudpy/gui/widgets/core/control.py
```diff
@@ -0,0 +1,627 @@
...
+        self.mainWidget.updateWidgets(self.gudpy.gudrunFile)
+        self.mainWidget.setWindowTitle(
+            f"GudPy - {self.gudpy.gudrunFile.filename}[*]")
+
+    def newProject(self):
+        if self.gudpy.gudrunFile:
+            save = QtWidgets.QMessageBox.question(
+                self.mainWidget,
+                "GudPy",
+                "Would you like to save current project?"
```
trisyoungs:
```suggestion
                "Would you like to save the current project?"
```

gudpy/gui/widgets/core/control.py
```diff
@@ -0,0 +1,627 @@
...
+                "Choose save location",
+            )
+            if not dirname:
+                return False
+            self.gudpy.setSaveLocation(dirname)
+            return True
+
+    def save(self):
+        # Check if save location has been set
+        # override project input file or force save location to be chose
```
trisyoungs:
```suggestion
        # Override project input file or force save location to be chosen
```

gudpy/gui/widgets/core/control.py
```diff
@@ -0,0 +1,627 @@
...
+        """
+        Exits GudPy - questions user if they want to save on exit or not.
+        """
+        self.cleanup()
+        result = QtWidgets.QMessageBox.question(
+            self.mainWidget,
+            "",
+            "Do you want to save before exiting?",
+            QtWidgets.QMessageBox.No | QtWidgets.QMessageBox.Yes,
+        )
```
trisyoungs: It looks like we always ask this - we know if the file has been modified though, right, so we can avoid asking it if the file is up-to-date.

gudpy/gui/widgets/core/worker.py
```diff
@@ -1,252 +1,235 @@
...
+
+    def _progressChanged(self):
+        stepSize = math.ceil(100 / len(self.dataFiles))
+        progress = 0
+        for df in self.dataFiles:
+            if df in self.output:
+                progress += stepSize
+        return progress
+
+    def appendDfs(self, dfs):
```
trisyoungs: Function name is a bit obtuse! Could we go for something more like:
```suggestion
    def appendDataFiles(self, dfs):
```

trisyoungs (approved):
Huuuuuge!

It's hard to do anything other than a sweeping review on a PR of this size, but the overriding impression is that this makes things a lot more harmonious, a lot cleaner, and a lot easier to read and understand workflow-wise compared to what we had before. Fabulous job!

Some small comments and typo corrections, but basically ready to go in and move on.

PR #493, opened 2024-03-20
Fix Purge Setting

rhinoella:
Some changes to fix the purge object passing were lost in https://github.com/disorderedmaterials/GudPy/pull/487, which prevents the purge results being accessed by Gudrun. This PR readds this change.


trisyoungs (approved):
:+1:

PR #498, opened 2024-04-18
Change all tabs to ScrollArea to accomodate for lower resolution screens

rhinoella:
GudPy is unusable on low resolution screens. This PR allows the window to be resized by setting a scoll area and minimum size.


trisyoungs (approved):
If you're sure you want to move to scroll areas, go for it!

PR #504, opened 2024-05-01
503 Fix Sample Tree Issues

### LINKED ISSUE #503: GudPy sample import fails
When importing a sample file it fails with the output:

```
Attribute Qt::AA_ShareOpenGLContexts must be set before QCoreApplication is created.
Designer: The enumeration-type property verticalScrollBarPolicy could not be read.
/home/nimrod_mgr/src/GudPy/gudpy/gudpy_gui.py:13: DeprecationWarning: 'exec_' will be removed in the future. Use 'exec' instead.
  sys.exit(self.exec_())
H-CSA
D-CSA
8mm_empty_quartz
HD-CSA
H-CSA+25mg/mL_HiPCO
Traceback (most recent call last):
  File "/home/nimrod_mgr/src/GudPy/gudpy/gui/widgets/core/gudpy_tree.py", line 290, in setData
    [s.name for s in self.gudrunFile.sampleBackground.samples],
AttributeError: 'GudrunFile' object has no attribute 'sampleBackground'. Did you mean: 'sampleBackgrounds'?
```

rhinoella:
Closes #503 

Fixes current issues with GudPy tree name checking, enabling/disabling samples and related bugs.


gudpy/gui/widgets/core/gudpy_tree.py
```diff
@@ -275,7 +275,7 @@ def setData(self, index, value, role):
         if not index.isValid():
             return False
         elif role == Qt.CheckStateRole and self.isSample(index):
-            if value == Qt.Checked:
+            if value == 2:
```
trisyoungs: Aside from the fact that magic numbers like this are bad, I'm not sure I understand what's being checked for here.

trisyoungs (commented):
Interestingly your CI is failing on one of the OSX builds because of stray quote marks in the Gudrun source.  I had the exact same thing on a Linux machine somewhere very recently and had to manually go and remove them (there are a couple of them)... migiht be something we need to address as a general issue.

gudpy/gui/widgets/core/gudpy_tree.py
rhinoella: To check whether the check box is checked or unchecked

rhinoella:
> Interestingly your CI is failing on one of the OSX builds because of stray quote marks in the Gudrun source. I had the exact same thing on a Linux machine somewhere very recently and had to manually go and remove them (there are a couple of them)... migiht be something we need to address as a general issue.

The issue is caused due to the github workflow update, amended in #506

trisyoungs (approved):
:+1:

trisyoungs (approved):
I think I'd already approved this, but I'll approve it again!

PR #505, opened 2024-05-03
Fix Output Tree

rhinoella:
Rewriting the output tree creation to smooth out the process and fix issues with iterator outputs.


trisyoungs (approved):
Cleaner!

:+1:

PR #510, opened 2024-05-22
feat: Update from v3 to v4 for uploading artifacts

rhinoella:


trisyoungs (approved):
Sensible to split the artifacts by OS.

:+1:

rhinoella:
> Sensible to split the artifacts by OS.
> 
> 👍

It is also a breaking change in V4- you can no longer upload artifacts to the same name

PR #513, opened 2024-05-29
feat: Include Gudrun dependencies in the run

rhinoella:
This PR makes sure that the necessary dependencies for gudrun are included in the running environment, such as the selfscatteringfile which is used by the inelasticity iterator.

It also serves to improve the output file management, instead of passing around a object to refer the run files, the outputs of the runs are stored within the dataFiles objects.


trisyoungs (approved):
Looks good.

:+1:

